PR #38, opened 2015-02-12
0.2.0

mottosso:
Version 0.2.0
-----------------

- Swapped controller from QML to Python
- Performance enhancements; the GUI now rivals non-GUI modes.
- Simplified terminal output; it now resembles non-GUI output
- Simplified backend; no more thread-per-process
- Temporarily switched to native OS window; this should help GUI appearing behind windows
- Temporarily disabled pause/stop buttons; to stop, close the GUI.


mottosso:
### Processing performance

For comparison; publishing from the current Pyblish QML is exponentially slower than in this version. The formula looks something like this.

```yaml
plugin: Time for plugin to process
polling: Fixed interval for polling between Maya and QML (20ms)
timeTaken: (plugin + polling) * numberOfPlugins
```

Which means that even if a plug-in took 0.002 ms to complete, which isn't uncommon, it would still have taken QML at least 20 ms to hear about its completion.

Whereas in `0.2.0` it looks like this.

```yaml
timeTaken: plugin * numberOfPlugins
```

No more polling means native speed; like when running `pyblish.main.publish()`.

### Bootup time

Before this release, booting up took around 3-4 seconds. It now takes ~0.8 seconds.

Issue #115, opened 2015-08-11
Validate


mottosso:
### Goal

Enable users to run validation only, to evaluate context without actually publishing.

tokejepsen:
Would this not just be having only validators for plugins?

mottosso:
This would run all Validators, even when you have other plug-ins available. It can be useful when you are only looking for a report while working.

Here's an example.

![untitled](https://cloud.githubusercontent.com/assets/2152766/9195786/78adbbac-401d-11e5-850d-2477750b61c2.gif)

tokejepsen:
Ahhh, I see.

BigRoy:
How does it know up to where to run the processes when using custom ordered plug-ins?

I think it might be nice to have the Flask button to represent a 'Run till order X' where right-clicking it gives you a drop-down menu of possible orders. Possibly highlighting the processes in the UI as you hover over the options.

This way it could run "Up to Validation", but also "Up to Report" if you have custom reporting plug-ins?

mottosso:
The plug-ins to be run are those that pass this test.

```python
import pyblish.api
from pyblish.plugin import inrange

class MyPlugin(pyblish.api.Plugin):
  order = pyblish.api.Validator.order + 0.1

if inrange(MyPlugin.order, base=pyblish.api.Validator.order):
  # run it
```

Validators always ranges from 1, plus/minus 0.5.

mottosso:
> This way it could run "Up to Validation", but also "Up to Report" if you have custom reporting plug-ins?

I don't really see a usecase for that. Orders aren't artist-facing, but highly internal to the development of plug-ins.

It also won't transfer to when we start using graphs. Orders are an implementation detail, for now.

BigRoy:
> It also won't transfer to when we start using graphs. Orders are an implementation detail, for now.

Fair enough. Just checking what the conditions are.

mottosso:
No problem Roy, it's good that you ask. Sorry to be short! A lot of juggling at the moment, trying to get things out of the way quickly. Perhaps a bit too quickly.

mottosso:
Implemented since a few versions back!

Issue #114, opened 2015-08-11
Configurable Client Ports


mottosso:
### Goal

Enable users to customise which ports should be used to communicate with Pyblish RPC.

### Motivation

Currently, clients launch at 6001 and increment from there. Next client occupies 6002, 6003 and so forth. However, it's likely that this series of ports may already be in use. As discussed in this [forum thread](http://forums.pyblish.com/t/publish-from-two-hosts-simultanously/126/59).

### Implementation

During host set-up, provide an option to customise this port.

```python
import pyblish_maya
pyblish_maya.setup(port=9001)
```

Which means integration in these cases cannot rely solely on environment variables; i.e. the supplied `userSetup.py` in case of Maya. But will instead need to provide their own tailored `userSetup.py`.

##### Globally Used Ports

If `psutil` is available, Pyblish also take into consideration ports in use by other processes on your local system. It is recommended that you install it, but may already be provided with your distribution of Pyblish, such as pyblish-win.

Otherwise, it's possible that you may publish from a few hosts, but eventually encounter a limit of possible Pyblish-enabled hosts open at the same time.

tokejepsen:
Can we not use environment variables for this?

tokejepsen:
Would be great to use a single environment variable ```PYBLISHPORT```, that will be used across all integrations.

mottosso:
I'll have a look.

mottosso:
Having trouble with this, it will have to wait until the next update.

The trouble is synchronising the two processes; your host and the Pyblish QML Python interpreter. Providing an environment variable when launching say Maya should "pass it on" to the Pyblish QML interpreter, but it doesn't look like it is, or I'm missing something else.

It's difficult to debug currently and simplified debugging is on the way. Until then, hope this will suffice!

Also, the default `first_port` got changed to `9001`, so you shouldn't need to do much.

tokejepsen:
> Also, the default first_port got changed to 9001, so you shouldn't need to do much.

Ahh, fair enough. That was what I needed here:)

PR #116, opened 2015-08-11
Validate

mottosso:
Primarily a maintenance update, with an added feature; validation-only.

See #114 and #115 for more.


BigRoy:
How does this relate to custom ordering of Plug-ins. How does it know to pass on only till Validation?
I think it might be nice to have the bottle represent a 'run till order X' kind of option.

Something like right clicking the bottle gives you a dropdown of Orders where as you hover over them the UI highlights up to where you will be running the processes.

Happy to see the button in there already. Sweet!

Issue #135, opened 2015-12-16
Quick Toggles


tokejepsen:
**Goal**

Quickly toggle multiple instances.

**Motivation**

Publishing with more than 3 instances that aren't wanted, and needing untoggling, can quickly become tedious. Especially when resetting the publish after validation, resets the toggle states.

**Implementation**

The idea is to have a ```toggle_all``` checkbox for each instance family, as illustrated below for the ```deadline.render``` family.

![d2bd9f146c94be1092e84c1466f45482584a8354_1_349x500](https://cloud.githubusercontent.com/assets/1860085/11847401/13549b84-a415-11e5-9fd2-da675be8192b.png)
![7ee4e1ebe756f4639f186ae221738cb2ee0c37be_1_349x500](https://cloud.githubusercontent.com/assets/1860085/11847400/13507356-a415-11e5-8cbc-b3cabf70df64.png)

mottosso:
Referencing this one so it doesn't get lost.
- https://github.com/pyblish/pyblish-qml/issues/26

Your one is better, because it has pictures. :)

PR #137, opened 2015-12-21
Quick Toggles working version

tokejepsen:


mottosso:
So clean! Haha, I was expecting loads of code for this feature, this looks great!

tokejepsen:
Thanks:)

For reference this is for #135

mottosso:
- https://github.com/pyblish/pyblish-qml/pull/137/files#diff-d1c4637cf11b5d8fc458e4f164effb6cR79

This is a good example of the differences between items in the model, and items purely in the view. You've had to set the visual state of this object in QML directly, which means no logic could easily be applied to it. For example, what if that button was not optional? What happens with the rest of the items, is that Python receives the signal to check that box, and validates whether it should actually proceed with actually unticking the visual state of that box. And only until then does it go ahead and untick it. Python unticks it.

In this case, Python is kept out of the loop, which is most probably fine as it's merely a convenience toggle. But just thought I'd take a moment to highlight a key difference here between MVC and a traditional "procedural" approach.

tokejepsen:
Yeah, I think this is why I was asking about the ```ProxyModel``` etc. to get the section delegate into Python realm, but this works fine:)

mottosso:
Yeah, looks great, did not consider that you could assign your own delegate as a Section!

mottosso:
Having a look at this now. One of the thing I was worried about when introducing a new item, the Section, into the Model on the Python-side, was how the processing would get affected. You would have needed to filter out those when traversing the model in the processing function; which incidentally is quite complex and fragile.

But since it's still based on the QML section, those problems never happened.

About the look.

image

I'm not convinced the indenting, and keeping it all to one side is the way to go.. It does look quite bloated with information.

What do you think about perhaps hiding the section checkboxes by default, and showing them on hovering?

tokejepsen:
> What do you think about perhaps hiding the section checkboxes by default, and showing them on hovering?

We could do. And in the same direction you could not have a checkbox by the sections, and let right-clicking on the section label trigger the toggling.
But my problem with this, is that users will have to know about the feature to use it or accidentally discover it, rather than spotting the checkbox and naturally explore its functionality.

mottosso:
> And in the same direction you could not have a checkbox by the sections, and let right-clicking on the section label trigger the toggling.

That could work actually! Perhaps the section could light up when it's being hovered, indicating that it could be clicked. In that case, it won't matter what "state" it was in before, the user could simply click it to see what happens.

I think "hiding" things in this way is ok; imagine we had twice the amount of features we have now, some would naturally be more important than others and those should ideally take the stage. Letting other, less important feature be discovered later on. The classic menu is a good example. It doesn't show you all of its items all of the time, but instead hints towards the availability of more items once clicked.

tokejepsen:
> That could work actually!

Alright, I'll explore this some more. Do we close the PR in the meantime?

mottosso:
No no, whenever you commit to your own repo, to the master branch, this PR will update to reflect that, and we can continue the conversation.

tokejepsen:
Have a look now:)

tokejepsen:
Quick screen recording; https://youtu.be/_kU8xlUuGgE

mottosso:
Oooo, that *is* nice. What do you think? :)

mottosso:
Clicked around a bit, perhaps the flashing to black is a bit hard on the eyes, what do you think about instead flashing to pure white? Leaving the hover color just a tad below white.

tokejepsen:
Was thinking that as well. This should be better.

mottosso:
I think that looks great. Let me know when you think you're finished, and I'll merge this.

tokejepsen:
I'm finished☺

PR #138, opened 2015-12-22
Cosmetics and stability

mottosso:
Updated compatibility test.

```bash
python -m pyblish_qml.compat
```

And refactored the QML SectionItem.

@tokejepsen There are two changes to the code you committed yesterday.

1. SectionItem became it's own file, next to other list items, just to make `Overview.qml` smaller and more readable.
2. The "imperative" code where things happen at an event are now "declarative" instead; this is the QML way and works due to the so called "bindings". Every attribute will automatically update itself once the expression changes, in this case, then the item is hovered or pressed. It's one of the differences between Python and QML and takes some getting used to, but it's really cool I think!

I've attempted to bring some of the binding-functionality into Python via the model; as you may have noticed, when you change any attribute on any of the items in the model, the UI updates. Traditionally, you would instead call upon a wrapper function of the model called `setData` for this to work.

That's where this little sucker comes in: [models.PropertyType](https://github.com/pyblish/pyblish-qml/blob/42309a4db31d2215fb5fc484e4c1ca389ea230c5/pyblish_qml/models.py#L102)

### DIFF
--- a/pyblish_qml/compat.py
+++ b/pyblish_qml/compat.py
@@ -76,9 +76,9 @@ def test_pyblish_availability():
     try:
         __import__("pyblish")
         __import__("pyblish_qml")
-        __import__("pyblish_endpoint")
+        __import__("pyblish_rpc")
     except:
-        raise Exception("Pyblish Suite not found")
+        raise Exception("Pyblish not found")
 
 
 def test_qtconf_availability():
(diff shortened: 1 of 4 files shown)


tokejepsen:
Testing this out, and the plugins section delegate are highlighting as well on hovering but have no functionality. Think this can be confusing for the user. I initially had two different delegates for instance and plugin sections.

mottosso:
Mm, I noticed this too. Better disable hovering and toggling unless an onClicked is assigned.

PR #140, opened 2015-12-22
Plugin section toggling

### LINKED ISSUE #139: Clicking section for plug-ins has no effect
In [a recent update](https://github.com/pyblish/pyblish-qml/pull/138), clicking sections was introduced. But only for instances. We'll need to either disable feedback for plug-ins, or add support there as well.

tokejepsen:
Fixes #139

### DIFF
--- a/pyblish_qml/qml/Overview.qml
+++ b/pyblish_qml/qml/Overview.qml
@@ -113,6 +113,11 @@ Item {
                 section.property: "object.verb"
                 section.delegate: SectionItem {
                     text: section
+
+                    onClicked: {
+                        checkState = !checkState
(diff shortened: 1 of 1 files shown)


mottosso:
Sweeeeet. :)

mottosso:
Would you mind if I delayed pushing this to Win immediately? It's a bit of a process to increment its version and produce another build and is better suited for collections of features than single ones such as this.

tokejepsen:
Yeah, I don't mind. I've got individual repos setup here, so I can just update whenever.

mottosso:
Ah, excellent!

Issue #142, opened 2015-12-22
Plugin file path in Perspective


tokejepsen:
Having multiple development environments can quickly become confusing, especially where plugins are located on disk.

Here are some options:

![basic_info](https://cloud.githubusercontent.com/assets/1860085/11954009/bf794686-a89c-11e5-894c-fa59bdcb1e4d.jpg)

![filepath](https://cloud.githubusercontent.com/assets/1860085/11954008/bf78fcb2-a89c-11e5-9d2d-b5f1c0314a9b.jpg)

If the plugin is in memory, this would be good to indicate in the path as well.

mottosso:
I know we talked about this somewhere.. good there's an issue for it, definitely something to add.

PR #143, opened 2015-12-22
Fix for #142

tokejepsen:


mottosso:
Hm, I think putting it right next to the for-loop would make it extra clear what's going on. What do you think?

tokejepsen:
I agree:)

mottosso:
Awesome, now let's have a look at some corner cases.

First of, in-memory plug-ins are all prefixed with `mem:`, so that's something we could improve.

image

Secondly, what happens when the path is too long? What should happen? I'm thinking it should crop to the width of the window and be selectable, we can do that by simply changing the component to one that supports selecting.

Btw, as a development hint, you can launch `app.py`, and then show Pyblish from for example Nuke. That way, you can show it and register custom plug-ins and such for debugging.

tokejepsen:
> First of, in-memory plug-ins are all prefixed with mem:, so that's something we could improve.

Change to ```In memory: Collector```?

> Secondly, what happens when the path is too long? What should happen?

Does it not just expand for lengthy paths?
capture

mottosso:
> Change to In `memory: Collector`?

Not sure, maybe something more subtle, like changing the color of it? I think you should be able to apply HTML to the text field. Something like `<font color="steelblue">Collector</font>`. Or maybe it's good as-is?

> Does it not just expand for lengthy paths?

It does, but I'm sceptical to how readable or useful that is.. especially since it can't get selected, which is what I'd suspect what most would expect to do after reading it, so it can be opened somewhere. Maybe let's keep that as a 1.0, and look at improving it once it's had some use?

tokejepsen:
> Or maybe it's good as-is?

Actually think its alright, if you are messing around with in memory plugins, you'll know not to expect a path for all plugins.

> let's keep that as a 1.0, and look at improving it once it's had some use?

Yeah, let's do that. I thought selectable text was a fix across the whole repos, along with hyperlinks.

mottosso:
> I thought selectable text was a fix across the whole repos, along with hyperlinks.

It's really only an issue with the listviews, like the terminal, due to the way it's implemented (well, as a list-view). But single label-like components, like this one, is just a toggle to enable. Including enabling of hyperlinks.

- [For example](http://doc.qt.io/qt-5/qml-qtquick-textedit.html#selectByMouse-prop).

Ok, when you're happy with this, feel free to merge!

tokejepsen:
I'll merge this now then:)

Issue #92, opened 2015-04-29
Default on/off option for Plugins


mkolar:
### Goal
Plugins need a way to choose whether they should be turned on or off when UI loads. This is to allow certain plugins to be available to artist if needed in certain occasions. 

#### Usecase
Following conform plugins might be used to integrate with ftrack, however we might only want artist to send a notification to supe in case of a milestone update to version (blocking finished, proposed final.)

* Publish version to ftrack (yes)
* Attach instaces as components (yes)
* Notify supervisor (false)

mottosso:
Thanks @mkolar 

About the interface, how about something like this?

```python
class MyPlugin(...):
   toggled = False
```

mottosso:
I've also been considering implementing flags to lessen the risk of getting an unwieldy amount of attributes available to plug-ins.

```python
# B
class MyPlugin(...):
   flags = pyblish.Toggled | pyblish.Optional
```

In the case of GUI-specific flags, this could potentially be distributed further. This could potentially act to separate between what is core-Pyblish, and what is relevant to any particular addon, like Pyblish QML.

```python
# C
class MyPlugin(...):
   flags = pyblish.Optional
   uiflags = pyblish.Toggled
```

For greater future-proof, we could potentially implement per-addon flags.

```python
# D
import pyblish.api as pyblish
import pyblish_qml

qml_flags = pyblish_qml.Flags()
qml_flags |= pyblish_qml.Toggled | pyblish_qml.Purple
core_flags = pyblish.Flags()
core_flags |= pyblish.Optional

class MyPlugin(...):
   flags = [core_flags, qml_flags]
```

Though it increases the amount of typing.

mkolar:
From point of view of readability of the code (hence being able to take someones plugin and start using it), I'd say that option A is the clearest. 

If really needed, then B to keep plugin development as simple as possible.

mottosso:
I agree, the only thing I'm not too fond of about A is how it couples an otherwise conceptual object with a specific graphical representation, in this case an item in Pyblish QML.

As we go, there will be more libraries that build upon the core library, and who knows, some might interpret `toggled` differently than this particular QML interface. At that point, we're in trouble.

More realistically, once Event, Docks, QML and future libraries start taking hold, they will each need more and more plug-in specific attributes. I'm still not sure how to best keep this separate from the core, but I think it needs to start soon.

mottosso:
With flags, at least we could introduce new flags on a library-basis.

```python
[...]

PR #146, opened 2016-01-06
#92

tokejepsen:
ValidateDefaultOff has the attribute ```active``` set to ```False``` which is now reflected in the UI.

capture

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,7 +1,13 @@
 Changelog
 =========
 
-Version 0.3.1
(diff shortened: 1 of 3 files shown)


mottosso:
Great work! Quick sanity check; does the plug-in get processed once you toggle it ON?

tokejepsen:
> does the plug-in get processed once you toggle it ON?

Yup:)

mottosso:
Win!

Issue #25, opened 2014-12-17
Hyperlinks in Output


davidmartinezanim:
I think that it would be nice if the output panel had some kind of rich formatting.

I don't think that we need anything overly complex. Having said that it would be great if any URLs are converted into hyperlinks that can be clicked (triggering default browser and such).

When a Publish fails, I want to provide the artist not only with a message but a link to a WIKI page explaining what needs to be done to get the plugin to succeed. This kind of training is specially useful for new starters that are not familiar with the pipeline.

I other words, it would help us *empowering* artists.

tokejepsen:
Heres a preview of a working version of this;

![capture](https://cloud.githubusercontent.com/assets/1860085/12151047/0085e59e-b4a5-11e5-841f-0e7e46301019.PNG)

The blue links are generated by html code in the text;
```html
<a href="http://pyblish.com/">click here</a>
```

And the white links are generate automatic from recognizing a "http" pattern.

mottosso:
Woot woot! Great work!

PR #147, opened 2016-01-07
fix for #25

tokejepsen:


mottosso:
Don't forget to include an image of what it looks like if it's a graphical change, so it's easier to evaluate.

tokejepsen:
Here we are:)

capture

mottosso:
Thanks. :)

I think the blue color is a bit sharp on the eyes against the gray background.. What do you think? Is that anything we could have a look at changing?

tokejepsen:
> What do you think? Is that anything we could have a look at changing?

My thoughts exactly, but that is pure html code in the text. This would be up to the user to change that. When the user just adds text based links, they will be converted to the white colour seen on ```http://pyblish.com```

mottosso:
I was thinking we could alter the default colors, via QML "Styles". A component you add to another component to govern how it looks

For example, here's how a default button can be modified.

- https://doc.qt.io/qt-5/qml-qtquick-controls-styles-buttonstyle.html

However, after some quick googling, it looks like there aren't any text components to modify the default colors of hyperlinks..

Since you're already doing string substitution, what do you think about defaulting the look of <a> tags to white as well? Basically ignoring whatever color is passed in (discouraging the use of multiple colors) and assuming none was present.

I think consistency here more expected than getting different colors for links you hard-code from those you don't.

tokejepsen:
> Since you're already doing string substitution, what do you think about defaulting the look of tags to white as well?

We could do, but I thinking that if someone is using html code there is a definite intent on something specific. It would then be very confusing if they get something they don't expect.

mottosso:
To be honest, I think the HTML tags in a Python docstring looks terrible and I think it's something we should discourage all together.

If you want a clickable link, just put http:// or https:// in front of it. That way, we still conform to docstring syntax and don't encourage further mutilation with more HTML tags, because that can get real ugly real fast.

tokejepsen:
Cool, don't need to do anything then:)

mottosso:
Haha, I suppose not. We'll simply not advertise the ability to use HTML. There is another reason for avoiding that too, which is that the Text component is severely limited in the amount of tags it supports. Once we start advertising HTML functionality, a lot of requests for missing functionality is bound to appear; better save that for something that actually supports HTML or roll our own parser (like Markdown-to-QML).

mottosso:
Kickass work @tokejepsen, I don't even recognize Pyblish anymore, it's so much better already. :)

tokejepsen:
Thanks:) got a couple of issues out of the way with some downtime here.

PR #148, opened 2016-01-14
Data Handlers

tokejepsen:
https://github.com/pyblish/pyblish/pull/238

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -356,7 +356,7 @@ class Controller(QtCore.QObject):
         item = self.item_model.items[source_index]
 
         if item.optional:
-            self.host.emit("instanceToggled", instance=item,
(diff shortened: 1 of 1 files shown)


mottosso:
A little cautious about the duplicated effort in toggling sections and toggling items alone, I would likely have the section toggling call upon the individual item toggling somehow. But we can deal with that later.

Looks good to me, any comments?

tokejepsen:
> A little cautious about the duplicated effort in toggling sections and toggling items alone, I would likely have the section toggling call upon the individual item toggling somehow. But we can deal with that later.

Yeah, not fond of it either, but it works:)

mottosso:
Success!

tokejepsen:
Excellent:) Gunna start rolling it out here asap.

Issue #134, opened 2015-12-15
Asynchronomise Refresh


mottosso:
### Goal

Hitting refresh currently involves a pre-processing step which on a slow/networked system produces a noticeable lag. Remove this lag.

This call is likely responsible for the lag.

- [self.host.reset()](https://github.com/pyblish/pyblish-qml/blob/8d897c343fc7f31c868a6680a11f84dbe25a19f0/pyblish_qml/control.py#L506)

It sends a request to the host to run `discover()` which can take a while, and it doesn't return control to the GUI until it has received a reply.

In fact, this whole block is a bit of a hog.

- [Block](https://github.com/pyblish/pyblish-qml/blob/8d897c343fc7f31c868a6680a11f84dbe25a19f0/pyblish_qml/control.py#L502-L510)

### Implementation

With #133, this step should be resolved.

Issue #133, opened 2015-12-15
Visualise which collector is running


mottosso:
### Goal

Before collectors were visible in the GUI, while they were loading there were nothing visual available to keep the user entertained. So, a spinning wheel was introduced.

Now that collectors are visible, it would be nice to see which collector is actually doing something, and which instances are being created, as they are being created.

### Implementation

Simply unhiding plug-ins as-is should reveal enough information to keep most users interested. But we'll also need to communicate partial information about which instances are available per process, and dynamically populate that list. Keeping in mind that this will have to happen asynchronously, such that the user can still interact with the GUI as it is running.

Issue #101, opened 2015-06-14
Plugin with family processing context shows in GUI even if family not present


BigRoy:
It seems that any `Plugin` restricted to a `family` that solely processes a context will always show up in the GUI.

```python
class ExampleValidator(pyblish.api.Validator):
    families = ['model']

    def process(self, context):
        pass
```

This Validator will always be present in the GUI even if no family of `model` is present in the context/instances. Note that it will not get run, so it seems to be solely a visual bug.

PR #150, opened 2016-01-20
0.4.0 - Maintenance update

mottosso:
This implements a number of issues and tweaks!

- Enhancement: Collectors visible during reset (#133)
- Enhancement: Animated feedback on currently running items
- Enhancement: More visible indicator of which is currently running
- Optimisation: Start-up performance; the Window pops up faster thanks for not having to wait for Collectors
- Optimisation: Interactivity performance; pressing buttons doesn't lag anymore (#134)
- Bugfix: Accurate representation of currently running items
- Bugfix: Both family and families are represented in the GUI
- Bugfix: Plug-ins only compatible with the Context are more accurately represented. (#101)

untitled

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,7 +1,20 @@
 Changelog
 =========
 
-Version 0.3.3
+Version 0.4.0
+-------------
+
+- Feature: Callbacks (#148)
+- Enhancement: Collectors visible during reset (#133)
+- Enhancement: Animated feedback on currently running items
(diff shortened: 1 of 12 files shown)


mottosso:
About this particular look; I've laid the groundwork for implementing progress bars. Anyone is welcome to modify the shapes and colors of this particular implementation. I'm not particularly keen on it as it is right now, but I'm heading to more critical portions of the API and expect this to be reasonably challenging for a contributor to accomplish.

Issue #125, opened 2015-09-15
Set a data member on toggle


tokejepsen:
When a user toggles the enable state in the UI, this only gets reflected in which plugins to process.

The idea is to set some data member on the asset, to indicate the enable state in the UI. Think the suggested data member was to modify the "published" data member, but I might be wrong @mottosso?

mottosso:
The toggled state will modify the `publish` data member, where `__published__` might become another variable indicating whether an `Instance` has already been published.

tokejepsen:
So to add further info for this, heres from gitter;

tokejepsen 12:05
So if I do a change to the context I get from self.host.context(), will this be reflected when publishing from the UI?
Or will it just get a fresh new context from the host prior to publishing?

mottosso 12:10
No, it's a one-way street. The context is created in the host, and sent across when you refresh.
To make a change, you'll need to build a new street, where information can go from QML to the host.
There was an implementation of this early on that never got finished, maybe it can be helpful here.
https://github.com/pyblish/pyblish-qml/blob/4d3a69778283b061f5697e3d3b7fc0a8ea815964/pyblish_qml/control.py#L428
The idea was that, whenever a particular item in the model changed, such as it's checked state, QML would emit this change to the host, which would in turn apply it to the instance or plug-in.
[This](https://github.com/pyblish/pyblish-rpc/blob/436ea351387c5159acb0fc06120d341dc97a6df7/pyblish_rpc/client.py#L18) is the object you access from self.host. You could add a method there called "save" which would call the remote host via [this](https://github.com/pyblish/pyblish-rpc/blob/436ea351387c5159acb0fc06120d341dc97a6df7/pyblish_rpc/service.py#L23). Then this is where you are basically inside of Nuke/Maya etc. and can modify the actual instances and plug-ins.

mottosso 12:17
You can think of [```pyblish_rpc.client```](https://github.com/pyblish/pyblish-rpc/blob/436ea351387c5159acb0fc06120d341dc97a6df7/pyblish_rpc/client.py#L18) and the [```pyblish_rpc.service```](https://github.com/pyblish/pyblish-rpc/blob/436ea351387c5159acb0fc06120d341dc97a6df7/pyblish_rpc/service.py#L23) as the Remote and Local borders, they have a 1-1 mapping between their methods and properties and handle the translation between the two borders.
.client runs in QML, and .service runs locally within the host.
You're entering the realm of IPC, it can be a bit daunting. 

tokejepsen 12:19
[...]

PR #155, opened 2016-02-03
fix for #125

tokejepsen:
- Fixing major bug of having switches new and old value around

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -8,19 +8,17 @@ from PyQt5 import QtCore
 import pyblish_rpc.client
 import pyblish_rpc.schema
 import pyblish.logic
-
+import pyblish.api
 
 # Local libraries
 import util
 import models
 
 from pyblish_qml import settings
-import pyblish.api
 
 
 def toggle_instance(instance, new_value, old_value):
     instance.data['publish'] = bool(new_value)
-    print instance.data
 
 pyblish.api.register_callback("instanceToggled", toggle_instance)
 
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -15,6 +15,14 @@ import util
 import models
 
 from pyblish_qml import settings
+import pyblish.api
+
+
+def toggle_instance(instance, new_value, old_value):
+    instance.data['publish'] = bool(new_value)
+    print instance.data
+
+pyblish.api.register_callback("instanceToggled", toggle_instance)
 
 
 def pyqtConstantProperty(fget):
@@ -408,8 +416,8 @@ class Controller(QtCore.QObject):
 
         if item.optional:
             self.host.emit("instanceToggled", instance=item.name,
-                           new_value=item.isToggled,
-                           old_value=not item.isToggled)
+                           new_value=not item.isToggled,
+                           old_value=item.isToggled)
 
             self.__toggle_item(self.item_model, source_index)
             self.item_model.update_compatibility()
@@ -422,8 +430,8 @@ class Controller(QtCore.QObject):
             if item.itemType == 'instance' and sectionLabel == item.family:
                 if item.isToggled != checkState:
                     self.host.emit("instanceToggled", instance=item.name,
-                                   new_value=item.isToggled,
-                                   old_value=not item.isToggled)
+                                   new_value=not item.isToggled,
+                                   old_value=item.isToggled)
 
                 item.isToggled = checkState
 
@@ -431,8 +439,8 @@ class Controller(QtCore.QObject):
                 if item.verb == sectionLabel:
                     if item.isToggled != checkState:
                         self.host.emit("pluginToggled", plugin=item.id,
-                                       new_value=item.isToggled,
-                                       old_value=not item.isToggled)
+                                       new_value=not item.isToggled,
+                                       old_value=item.isToggled)
 
                     item.isToggled = checkState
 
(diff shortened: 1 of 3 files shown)


tokejepsen:
Using a callback to set the ```publish``` data member in the host. Don't know how clean that is, or whether its in the right place.

mottosso:
Could you have a look at consolidating where this event is emitted, rather than doing it in multiple places? Something can easily be missed as it is now I think. Maybe the section toggle could make use of the instance and plug-in toggle methods somethow?

pyblish_qml/control.py
```diff
@@ -17,6 +17,12 @@
 from pyblish_qml import settings
 
 
+def toggle_instance(instance, new_value, old_value):
+    instance.data['publish'] = bool(new_value)
+
+pyblish.api.register_callback("instanceToggled", toggle_instance)
```
mottosso: I'm not a fan of having modules produce side-effects on import. How about putting this in the [main](https://github.com/tokejepsen/pyblish-qml/blob/1be49764f34605d532b68d600f097e6275475a08/pyblish_qml/app.py#L226) instead, so we can safely import it without worry?

tokejepsen:
Have a look at this now.

I'm slightly unsure of the "instanceAdded" callback solution, but its working.

mottosso:
Having a look..

mottosso:
Ah, the instanceAdded signal is currently coming from an internal class of the GUI. It doesn't have anything to do with Pyblish or publishing in general, but is just when the items are added visually to the GUI.

Did you have a usecase for this signal? It would find a better home in the Context class I think, possibly in thet `create_instance` method. But we shouldn't add signals unless there's a clear use for it. What do you think?

Otherwise, great work on the consolidation! It looks super smooth. I can't test it now, but by just eyeballing it I'd say it looks very maintainable and minimal.

tokejepsen:
> It would find a better home in the Context class I think

Not quite getting where this Context class is? In pyblish-base?

mottosso:
Yeah, sorry, I should have linked.

- [Context.create_instance](https://github.com/pyblish/pyblish-base/blob/a70c91c4620b5b4cb51cf749dc0de31f8fed909e/pyblish/plugin.py#L653)

tokejepsen:
> Yeah, sorry, I should have linked.

But the point was to stay away from the pyblish-base repo, right? Only add the ```publish``` data member from pyblish-qml (some how), because it is needed here.

mottosso:
> Only add the publish data member from pyblish-qml (some how), because it is needed here.

But this event isn't adding the `publish` data member? I thought it was just emitting whenever an instance got added to the Context?

Maybe let's start from the beginning. :) What is this event meant to do, and what is it for?

tokejepsen:
> But this event isn't adding the publish data member? I thought it was just emitting whenever an instance got added to the Context?

Ohh, right yea. If we are just emitting the signal, we can receive it when using pyblish-qml, meaning nothing gets added unless the signal is received. I get it:)

mottosso:
..now I don't get it. :)

Tell me, why this signal? It wasn't there before, why bother adding it?

tokejepsen:
> Tell me, why this signal? It wasn't there before, why bother adding it?

To add the ```publish``` data member to all the instances; https://github.com/pyblish/pyblish-base/pull/244#issuecomment-179720959

mottosso:
Ahaa, now I get you, you still want that to be there. I thought we agreed to not put data members in instances unnecessarily.

I think this is fine.

```python
if instance.data.get("publish", True):
    print "Instance is publishable"
```

What's wrong with that?

tokejepsen:
> What's wrong with that?

I think its confusing for new users, that you set the ```publish``` data member to ```False``` when untoggling by default, and then when you query for it, the data member doesn't exist.
I don't mind using this workflow, but I just recall my first intuition to query the toggle state of an instance in the GUI, and it was to look at the ```publish``` data member, which was why I issued #125.

mottosso:
Ah, ok. Sorry, I'm a bit slow today. :)

I can see what you mean, but let me know what you think about this, as this is the source of inspiration for I'm imagining things.

In Sublime Text, you have a default json configuration file for each feature, and an optional user-defined json file with overrides. The user-defined json file can contain every key in the default file, but only ever needs to contains what differs. In general, it only has one or two keys, whereas the default one has tens if not hundreds of potential targets.

I kind of like that workflow. It means they can change their default around, and your changes are just "layered" on top of those.

I've seen at least a few programs do things similarly, and it's how the configuration of Pyblish (when we were using json files to configure paths and such) was also working. But it does have the disadvantage of having users add entirely new keys to their user config, which can understandably be daunting.

I can see it both ways, what are your thoughts?

tokejepsen:
> I can see it both ways, what are your thoughts?

I'm probably not entirely following you trail of thought. This "layered" approach sounds good, but the problem is we don't have a default data member for the toggle state of an instance in the GUI, so there is nothing to override.
This is actually what confused me to begin with, that you add a data member for the GUI to recognize the default toggle state of an instance. I know now that we are making an assumption, but my first thoughts were; "How does the GUI know which toggle state an instance is in, if there is no data member to begin with?"

mottosso:
Let me have a think on this. The default I'm imagining would come from documentation, but there isn't any yet. Something like [this](https://github.com/pyblish/pyblish.api/wiki/Plugin). Listing data the GUI is looking for, along with what it considers default.

PR #160, opened 2016-02-07
Merge tokejepsen commit

mottosso:
@tokejepsen I've merged your fixes with some of the changes I've made and fixed a small typo in how `instanceToggled` was emitted (`name` instead of `id`).

I also removed signal emission from `model.py`. As we [discussed](https://github.com/pyblish/pyblish-qml/pull/155), events happening within the QML model are internal to the GUI and doesn't necessarily reflect useful events to processing of any plug-in.

Also, it was quite a nasty dependency; having the model implicitly communicate with the host. The controller is the only part supposed to bring the model and view together, along with handling IPC communication with the host. That way, both the view and model can remain blissfully ignorant, making it a bit easier to track down problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
 [![Build Status](https://travis-ci.org/pyblish/pyblish-qml.svg?branch=master)](https://travis-ci.org/pyblish/pyblish-qml)
-[![PyPi](https://badge.fury.io/py/pyblish-qml.svg)](http://badge.fury.io/py/pyblish-qml)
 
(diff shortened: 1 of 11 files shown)


mottosso:
Ok, I had a look at your usecase of having the `publish` data member apply to the current instances in the Context.

The way [the callback was registered now](https://github.com/pyblish/pyblish-qml/commit/ac667a1f8f74a63abb3aab578725961890f7d816), they only existed in the pyblish-qml Python process. They weren't actually present in the host, and therefore couldn't have any effect on the instances there.

I've moved these to pyblish-integration where it works as expected.

PR #164, opened 2016-03-02
Test for collection sorting

tokejepsen:


mottosso:
Great work! Can you think of a way to pass this test?

tokejepsen:
We would need to reorder the items in the item model, and refresh the GUI.

Unless you have any other ideas or comments, I'm going to look into this now.

tokejepsen:
I'd imagine calling a method at the end of the collection, to refresh the GUI order.

mottosso:
Perhaps, yes.

At the moment, instances are being added as they are created. We'd need to somehow delay adding them, or as you say, reorder them afterwards. I'm not sure which is easier. It's possible that you can simply reorder the list in the model, and emit the corresponding signal so it updates the GUI.

Do you know which signal I'm talking about?

tokejepsen:
> At the moment, instances are being added as they are created. We'd need to somehow delay adding them, or as you say, reorder them afterwards. I'm not sure which is easier. It's possible that you can simply reorder the list in the model, and emit the corresponding signal so it updates the GUI.

Looking into reordering them afterwards. If it'll be a problem for someone, we'll just have to delay adding them.

> Do you know which signal I'm talking about?

No. ATM I'm looking at reordering the item model, and it seems to update the GUI.

mottosso:
> No. ATM I'm looking at reordering the item model, and it seems to update the GUI.

Ah! That's great. I assumed you would need to let the model know you've changed it's internal structure.

Here's one of the signals to use for that.

- http://doc.qt.io/qt-5/qabstractitemmodel.html#beginMoveColumns

The other one is the "end" equivalent. It might be safest to do this, even if it looks like it works now. It really shouldn't be able to tell when you move things, unless you explicitly tell it you are. It might be a fluke at the moment.

mottosso:
Since the whole model is changing, you could probably get away with resetting the whole thing!

- doc.qt.io/qt-5/qabstractitemmodel.html#beginMoveColumnsh4

tokejepsen:
There is a corresponding mocking test, which I used when testing; https://github.com/pyblish/pyblish-rpc/pull/18

mottosso:
Awesome! When you're ready, and you feel it works ok, feel free to merge this.

tokejepsen:
Testing on a heavy production scene, successfully:) Merging.

mottosso:
The best kind of test. :) Thanks Toke!

PR #166, opened 2016-03-29
Action icons

tokejepsen:
**Goal**

To indicate where there are actions for the user to execute.

**Implemetation**

Whether a plugin has actions for the user to execute is indicated with an icon.
icon

The functionality of when this icon appears is working. For example the icon only appears when the action is ready, which could be in the case of failing plugins;
pre_validate
post_validate

What I'm missing is to color the icon depending on the success of the execution. The placement for changing the color is indicated in the ```control.py```, but color of the icon needs to be white to begin with, then colored after the action's execution.
If you execute one action after another, the color of icon just takes on the most recent execution.
red_icon
green_icon


BigRoy:
This is looking great! Thanks! 

The coloring would indeed be a nice addition.

tokejepsen:
Updated with action icon colouring, depending on execution status.

Also if you didn't notice I sneakily added colour to the warning icons as well.

mottosso:
> Also if you didn't notice I sneakily added colour to the warning icons as well.

Once you're happy with the pull request, I'll be sure to comment on things. :)

tokejepsen:
Comment away! As far as I'm concern its done.

mottosso:
I don't like the color of the warning icons. :)

The only colors I'm interested in, is those that communicate something. Red for failed, green for success. All else should lie silent in the background.

tokejepsen:
> I don't like the color of the warning icons. :)

Fair enough, I can revert that. Anything else?

mottosso:
Okay, sounds good.

How should we handle future default actions for plug-ins? For example, it would be nice with an action to open the parent directory in a browser, and another action to run a plug-in in isolation.

tokejepsen:
> How should we handle future default actions for plug-ins?

Not entire sure what you mean?

Currently the icon isn't a button, so people are still to right-click and select the action they want to exectute.

mottosso:
I mean the icon shows up if a plug-in has actions, but what about once every plug-in has at least 1 or more action? The icon would show up for every plug-in, which might not be very helpful.

tokejepsen:
What I could explore is to add the icons of the actions, instead of just having a single icon to indicate is a plugins has actions?

mottosso:
Yeah, that might be cool.

We'll have to be careful not to pollute what's important. Maybe show a few, and add a "hamburger" icon for the rest? Maybe the ones that are being shown can be shown based on some form of priority;

```class
class MyAction(...):
   priority = 2
```

That way, default plug-ins would fade into the background, and when a default plug-in really wants attention, it can boost it's priority. The same could go for user supplied actions.

tokejepsen:
> Maybe show a few, and add a "hamburger" icon for the rest?

Or ```...```?

> Maybe the ones that are being shown can be shown based on some form of priority

With the actions of a plugin being a list, could we not just have it be the first ones in the list?

mottosso:
> Or ...?

Possibly, but there's chance a hamburger can be made smaller horizontally than dots, while still being clear. Give it a go, try getting a feel for it.

> With the actions of a plugin being a list, could we not just have it be the first ones in the list?

Could start off with that. Just trying to think ahead.

tokejepsen:
I've had a brief look into this, and it seems quite a bit outside my QML knowledge, so it'll take me quite a while to sort out.

Could we merge this, and have a separate issue for multiple icons?
I'm currently trying to update our production pipeline, but this along with some other issues are standing in the way.

mottosso:
Sorry I'm not comfortable merging this in a rush.

I'll try take a look at this tonight, we'll need to get some tests in there as well.

BigRoy:
Having multiple icons shift to visibility on the Plug-in sounds even better than the (A) popping up. (E.s. just the first two/three). 

Or something I mentioned before on the forums like: `visible_icon = True` on the action instead assuming the order? The actions with that set to True would become visible on the plug-in when actions are available (e.g. on failed).

mottosso:
@tokejepsen with merge just now, there might have been some conflicts introduced between this PR and that one. Are you able to resolve these? You should be able to rebase.

If you're using SmartGit, after you've started the rebase, it will stop and let you resolve. Right click on the conflicting file and it will let you choose whether to keep your parts, or those recently merged. You can do the same via the terminal if you use plain git.

Let me know if you need help with that.

tokejepsen:
@mottosso I've updated the PR with the latest changes, and checked everything is working as before. I've also reverted the warning colour to white.

mottosso:
That's great! I'll have a play with this tonight and hopefully we can merge this then. Is that ok?

tokejepsen:
Were you going to look at the multiple icons?

mottosso:
No I think we'll start with this and open a new PR for that, as you said.

tokejepsen:
Should we make an issue for it, to remember?

mottosso:
Yeah, let's.

mottosso:
> Should we make an issue for it, to remember?

Did we expect each other to make this issue? :) I can't find it.

tokejepsen:
Thought #170 was it?

mottosso:
Ah, so it is. :) Now there's a link between the two so they won't get lost. Thanks.

PR #174, opened 2016-04-13
Fix #173

### LINKED ISSUE #173: Toggle instance doesn't  update compatibility view correctly
When multiple instances are present and you toggle the state the GUI doesn't update correctly. It seems to update to the previous state. Toggling multiple times always shows the plug-ins that were "compatible" with the previous state.

This seems to be related to https://github.com/pyblish/pyblish-qml/blob/master/pyblish_qml/control.py#L509

```python
        self.item_model.update_compatibility()
        item.isToggled = new_value
```

The state is set after updating the compatibilty. Switching these around fixes the problem.

```python
        item.isToggled = new_value
        self.item_model.update_compatibility()
```

Will push a fix later today.

BigRoy:
- Includes a test that fails with previous code and passes with fixed code

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -527,8 +527,8 @@ class Controller(QtCore.QObject):
                            new_value=new_value,
                            old_value=old_value)
 
-        self.item_model.update_compatibility()
         item.isToggled = new_value
+        self.item_model.update_compatibility()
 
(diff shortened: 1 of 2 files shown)


mottosso:
Great work Roy

tokejepsen:
Great job!

Issue #172, opened 2016-04-06
Actions Availability


tokejepsen:
If you have an action working on failed instances, the appearence of when you can right-click and select the action isn't consistent.

Here is a test for this case:

```python
import pyblish.api
import pyblish_integration


class SelectMany(pyblish.api.ContextPlugin):
    order = pyblish.api.CollectorOrder
    def process(self, context):
        for name in ("A", "B", "C"):
            instance = context.create_instance(name)
            instance.set_data("family", "myFamily")


class FailureAction(pyblish.api.Action):
    on = "failed"


class ValidateFailA(pyblish.api.InstancePlugin):
    order = pyblish.api.ValidatorOrder
    families = ["myFamily"]
    actions = [FailureAction]
    def process(self, instance):
        assert str(instance) != "A"


class ValidateFailBC(pyblish.api.InstancePlugin):
    order = pyblish.api.ValidatorOrder
    families = ["myFamily"]
    actions = [FailureAction]
    def process(self, instance):
        assert str(instance) == "A"


pyblish.api.deregister_all_plugins()
pyblish.api.register_plugin(SelectMany)
pyblish.api.register_plugin(ValidateFailA)
pyblish.api.register_plugin(ValidateFailBC)
pyblish_integration.show()
```

The test basically shows that the first instance in the execution list, is the determining factor for whether the action is available or not.
In the test the action should be available for both validators, but is only available for ```ValidateFailA```, because it fails the first instance in the execution list.
It does not have the action available for ```ValidateFailBC```, because instance ```A``` succeeds and instance ```B``` and ```C``` fails.

BigRoy:
Was about to post about this as it has been confusing both the artists (and myself!) as to why and when this was happening. Thanks for the brief explanation and pinpointing where the issue was coming from.

Any idea where to look in the code to get this fixed?

mottosso:
Here's the if-statement in control over when actions are visible and not.

- https://github.com/pyblish/pyblish-qml/blob/90db447049b17143283c60f08397f439ec4be236/pyblish_qml/control.py#L321

BigRoy:
Was looking there but couldn't really find (or debug) where the root of the proble was. Anyway, would be great to have this confusing issue solved.

tokejepsen:
Sorry, been trying to get some time to fix this, but it's running away from me. Got kinda stuck on getting a quick setup for restarting qml for development.

mottosso:
[...]

PR #176, opened 2016-04-15
fix for #172

tokejepsen:
Fix for #172

### DIFF
--- a/tests/test_control.py
+++ b/tests/test_control.py
@@ -430,7 +430,7 @@ def test_toggle_compatibility():
 
 
 @with_setup(lib.clean)
-def test_toggle_compatibility():
+def test_action_availability():
     """failed plugins with failure action, needs to have an action"""
 
     class SelectMany(pyblish.api.ContextPlugin):
@@ -469,5 +469,5 @@ def test_toggle_compatibility():
 
     validate(c)
 
-    assert c.item_model.plugins[1].actions
-    assert c.item_model.plugins[2].actions
+    assert c.item_model.plugins["ValidateFailA"].actions
+    assert c.item_model.plugins["ValidateFailBC"].actions
--- a/tests/test_control.py
+++ b/tests/test_control.py
@@ -378,14 +378,14 @@ def test_gui_vs_host_order():
     msg += "\n%s >> Host order" % host_order
     assert host_order == gui_order, msg
 
-    
+
 @with_setup(lib.clean)
 def test_toggle_compatibility():
     """toggle instance updates compatibility correctly"""
 
     class Collector(pyblish.api.ContextPlugin):
         order = pyblish.api.CollectorOrder
-        
+
         def process(self, context):
             instance = context.create_instance("A")
             instance.set_data("family", "FamilyA")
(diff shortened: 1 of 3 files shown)


mottosso:
Nice catch.

It looks like this runs, even when the icon is already visible. Since it runs once per result, and is a relatively complex loop, how about something like this before it?

```python
if item.actionsIcon:
  continue
```

That way, it only happens once.

Also, let's give `actionsIcon` a better name. It sounds like the icon for actions, when it acts like an indicator about whether it should be visible.

Does `actionsIconVisible` better state its intent?

Also, let's get a test in there as well. ;)

tokejepsen:
> Also, let's get a test in there as well. ;)

Questions is what I'm testing for?
Technically everything was working correctly, that problem was that the actions list got emptied.

mottosso:
Test for the expected behaviour.

The idea is that, whenever this breaks again, it will be caught by a test.

tokejepsen:
There we are:)

mottosso:
Almost got it!

This test would pass even if it didn't fail and there was an automatically added action, like a default action.

We need to make sure it doesn't pass when the problem is there, and try and limit the range of the test to the tiny portion it concerns.

Indexing into the `.plugins` list is fragile too; if Pyblish or a plug-in had decided to rearrange the registered plug-ins after they registered, this test would also fail; even though the thing it's testing should have been unaffected.

Using `.plugins["MyPlugin"]` is safer.

tokejepsen:
I thought the tests were isolated from other tests/plugins?

But yeah, I'll have a look at getting the plugins explicitly.

mottosso:
> I thought the tests were isolated from other tests/plugins?

How do you mean? If you're registering a plug-in in the same test, then it's safe to refer to explicitly. Just make sure you de-register them when you're done, to not cause side-effects for other tests. If you're unsure about how to do that that, use a unique name for the plug-in.

tokejepsen:
> Just make sure you de-register them when you're done, to not cause side-effects for other tests.

Alright, thought for some reason that the decorator took care of any cross test contamination, so the tests were isolated from each other.

mottosso:
Ah, yes you are right, Didn't notice you were using it!

tokejepsen:
I've updated the test to explicitly getting the plugins by name.
Also changed the name of the test, as it was the same as the one I copied from :worried:

BigRoy:
Wanted to say thanks again for jumping onto this! :)

By the way, looking at the test. Should it maybe also include a case where it doesn't have any actions?

mottosso:
Yes, thanks a lot!

> Should it maybe also include a case where it doesn't have any actions?

I think so too. I'm having a look at a few more tests for this now.

tests/test_control.py
```diff
@@ -402,28 +402,72 @@ def process(self, instance):
...
+    pyblish.api.deregister_all_plugins()
+    pyblish.api.register_plugin(SelectMany)
+    pyblish.api.register_plugin(ValidateFailA)
+    pyblish.api.register_plugin(ValidateFailBC)
+
+    c = reset()
+
+    validate(c)
+
+    assert c.item_model.plugins["ValidateFailA"].actions
```
mottosso: I've had a closer look at this, and I'm not sure this is right. The plug-in should always contain it's actions, it's only when asking for them that the filtering takes place.

- https://github.com/pyblish/pyblish-qml/blob/87cf2c8dfe500e5dd41d9d6c8e0005a45f3445ff/pyblish_qml/control.py#L302

tokejepsen: The problem wasn't in the controller, but in the model; https://github.com/pyblish/pyblish-qml/blob/87cf2c8dfe500e5dd41d9d6c8e0005a45f3445ff/pyblish_qml/models.py#L431

The problem was that the it didn't maintain its original actions, because the model emptied the actions list by not copying the list.

mottosso: But was the test correct as was merged just now? It would have passed regardless of the `on` state I think.

tokejepsen: > But was the test correct as was merged just now?

It failed with previous code, and passed with the fixed code. Or maybe I don't understand what you are asking?

mottosso: It passed because it would have passed regardless of a plug-in failing.

That list contains all actions of the plug-in, not the currently available actions in the GUI, that's what the [`getPluginActions()`](https://github.com/pyblish/pyblish-qml/blob/29c24ddbafaf2b6edc365ee3027596c580dd0e08/pyblish_qml/control.py#L302) method is for. That will take into account whether anything has failed or not.

I modified your test, and added a plug-in that did not fail. Have a look and see if it makes sense.

- https://github.com/pyblish/pyblish-qml/blob/master/tests/test_control.py#L487

tokejepsen: Cool, yeah, looks good to me.

PR #178, opened 2016-04-18
- fixing qml typo

tokejepsen:
Quite a huge mistake which meant Action icons weren't showing at all.

Clearly my test was not good at all :disappointed:


mottosso:
Nice catch!

This reminds me that we also need tests for QML, though I've never written any before. It's a rather new mechanism overall.

Here's the docs for completeness.

- https://doc.qt.io/qt-5/qtquick-qtquicktest.html

PR #181, opened 2016-04-27
Work in progress of new ID-based items

mottosso:
GUI works, but some tests fail!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,18 @@
 language: python
+
 python:
 - 2.7
+
+sudo: required # install.sh requires sudo apt-get on Linux
+
+install:
+- echo "Install.."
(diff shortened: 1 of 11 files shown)


BigRoy:
Thanks a ton. Is it mainly test changes from this point?

BigRoy:
Will have a quick look on my end with this.

mottosso:
> Is it mainly test changes from this point?

Possibly, but we're probably not out of the woods yet. The tests might reveal a flaw somehow.

BigRoy:
It was purely on IDs -> names, not something else. See my [PR](https://github.com/mottosso/pyblish-qml/issues/1). Ran this version through a quick scene as well and seems to work (just quickly tested) including the actions.

PR #182, opened 2016-04-27
Update server.py

linez69:
ping()


mottosso:
Great job!

For everyone else, this function can be used to test for an already-running Pyblish QML.

```python
import socket
from pyblish_qml import client

proxy = client.proxy()

try:
    reply = proxy.ping()
    print(reply)
except (socket.error, socket.timeout):
    print("No reply, Pyblish QML must not be running")
```

tokejepsen:
Awesome @linez69 :smile:

PR #192, opened 2016-05-06
making checkbox look more like real checkboxes.

larsbijl:
bumping the version


mottosso:
Merging this.

untitled

PR #197, opened 2016-05-15
basic pip setup

tokejepsen:
I'm not entirely sure what we should be including in the setup, so I've just stripped down the one from pyblish-base.

This is to be able to do a pip install from the repository with conda.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,13 +24,16 @@ version = version_mod.version
 qml_dir = os.path.abspath('pyblish_qml/qml')
 qml_package_data = list()
 for root, dirs, files in os.walk(qml_dir):
-    for suffix in ("ttf", "qml", "js", "txt", "png", "py", "otf"):
+    for suffix in ("ttf", "qml", "js", "txt", "png", "py", "otf", "ico"):
         relpath = os.path.relpath(root, qml_dir)
         relpath = relpath.replace("\\", "/")
         qml_package_data.append("qml/" + relpath.strip(".") + "/*." + suffix)
 
 # qmldir file has no suffix
 qml_package_data.append(os.path.join("qml", "Pyblish", "qmldir"))
+qml_package_data.append(os.path.join("qml", "Pyblish", "Graphs", "qmldir"))
+qml_package_data.append(os.path.join("qml", "Pyblish", "ListItems", "qmldir"))
+qml_package_data.append(os.path.join("qml", "Perspective", "qmldir"))
 
 classifiers = [
     "Development Status :: 5 - Production/Stable",
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,4 @@
-from setuptools import setup, find_packages
-
+from distutils.core import setup
 
 import os
 import imp
@@ -8,14 +7,10 @@ version_file = os.path.abspath("pyblish_qml/version.py")
 version_mod = imp.load_source("version", version_file)
 version = version_mod.version
 
-
 setup(
-    name="pyblish-qml",
+    name='pyblish-bumpybx',
     version=version,
-    long_description=readme,
-    url="https://github.com/pyblish/pyblish",
+    packages=['pyblish_qml',],
     license="LGPL",
-    packages=find_packages(),
-    zip_safe=False,
-    classifiers=classifiers
+    long_description=open('README.md').read(),
 )
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ version_mod = imp.load_source("version", version_file)
 version = version_mod.version
 
 setup(
-    name='pyblish-qml',
+    name='pyblish_qml',
     version=version,
     packages=['pyblish_qml',],
     license="LGPL",
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ version_mod = imp.load_source("version", version_file)
 version = version_mod.version
 
 setup(
-    name='pyblish_qml',
+    name='pyblish-qml',
     version=version,
     packages=['pyblish_qml',],
     license="LGPL",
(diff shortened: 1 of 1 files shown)


mottosso:
Ah, brilliant. I've had plans for getting all the pyblish projects back up to PyPI. 

Have a look at how pyblish-base manages to get this automatically uploaded to PyPI on each new release as well. You can test the install with pip and github, something like this.

```bash
pip install git+git://github.com/tokejepsen/pyblish-qml.git
```

Have you done that to make sure it installs ok?

tokejepsen:
Sorry, yeah was a bit too quick on the PR. Have updated with commits so it installs ok, and its been tested.

Let me know if I need to include anything else.

tokejepsen:
> Have a look at how pyblish-base manages to get this automatically uploaded to PyPI on each new release as well.

Is that to do with travis CI?

mottosso:
> Is that to do with travis CI?

Yes, Travis can trigger an upload on a new release.

I would suggest you familiarise yourself with PyPI first; create an account and upload a test project (or this project) that you can delete once you feel comfortable. PyPI is picky about what you upload and it won't let you change what you upload; that is, each upload is immutable. But it will let you delete and re-upload.

Once you've got something up on PyPI, automating it on Travis will make more sense. It involves setting the username and password, which I can provide, and generating a "token" (an encrypted password that reside in public on GitHub). I would suggest familiarising yourself with that as well, using your own GitHub and PyPI account, this can also be a bit daunting, but it's a healthy one-off you can use for any number of projects once  you're comfortable; and it'd be great personally having someone else knowing the ropes on this.

Up for the task?

mottosso:
By the way, you can install using pip from a file on disk as well; no need to upload to GitHub each time. It think it's simply prefixing the name with `file://`.

```bash
pip install file://path/to/pyblish-qml
```

tokejepsen:
> Up for the task?

Maybe at some point. At the moment I'm just looking to be able to install all the pyblish packages with pip through conda. I'm still not 100% sure conda is the right path for me, so I don't want to get sidetracked right now.

> By the way, you can install using pip from a file on disk as well; no need to upload to GitHub each time. It think it's simply prefixing the name with file://

Thanks, that'll make it quicker :smile:

tokejepsen:
@mottosso do you know what the problem is with the checks?

mottosso:
Yeah, don't worry about those yet. I started setting up this project for Travis testing, but haven't finished.

The other PR has Python 3 support which simplifies installing PyQt5 on Travis. It's just a matter of merging that and setting that up in the `.travis` file.

tokejepsen:
> Yeah, don't worry about those yet. I started setting up this project for Travis testing, but haven't finished

Cool, any problems with the ```setup.py```?

mottosso:
Hm, yes I'm not sure this actually works.

Are you sure you've tested it? I would expect the QML files, images and anything not Python to be excluded from the resulting pip installation. There is a special field for those in the `setup()` function. You'll need to scan the project for additional files, called "package data", and add their absolute paths to the function.

See here.

- https://pythonhosted.org/setuptools/setuptools.html#including-data-files

mottosso:
This project once had PyPI support. You can look at the previous setup.py for reference. It might work as-is, but I'd test it first.

- https://github.com/pyblish/pyblish-qml/blob/486f1e27e89b6fc8a13fee1853bf2374b3d1845a/setup.py

tokejepsen:
> Hm, yes I'm not sure this actually works.

It works with ```pip install file://path/to/pyblish-qml```, or do I have to test it some other way?

mottosso:
> or do I have to test it some other way?

Aah, yes I think it is misleading you.

Packaging for PyPI (and effectively GitHub as well) runs through a different process, one that selectively packages the files you point it at. That's what we need to test for in this case. I think it happens when you pip install via git, and I think you can do that locally as well, just point it at your local git repo.

If that doesn't reveal the problem, then try running this.

```bash
python setup.py sdist
```

It will create an archive containing all packaged files.

Reference.

- https://pythonhosted.org/setuptools/setuptools.html#basic-use

tokejepsen:
I've installed it locally with git and I can see that the qml files are indeed not coming along. How would I test that everything is working fine? Run ```app.py``` in debug mode, or run a some kind of other test?

mottosso:
> How would I test that everything is working fine?

Run the test-suite, though that won't test whether QML is working as it has no tests.

Best thing you can do, is run it and visually confirm that nothing is missing. If you use the previous setup.py I posted, you'll see that there aren't that many file-types of interest so if it even runs it's quite likely it's all there.

mottosso:
You could do something like this.

1. Collect Python package from original project
2. Collect Python package from pip generated distribution
3. Compare the two.

They should match exactly. You could write a test for that even.

tokejepsen:
Think I need some help with this. I think I have tried every possible combination to get the files to come across, but nothing seem to work. I've also looked at python-qt5 for inspiration.

mottosso:
Ok, having a look.

tokejepsen:
Got it working now! ```pip install pyblish-qml``` gives me the correct files etc.
I noticed that installing ```python-qt5```with pip doesn't work with this version of ```pyblish-qml```, which I guess is because its an old distrubution;
```
file:///C:/Users/admin/Documents/pipeline/miniconda/envs/pipeline/lib/site-packa
ges/pyblish_qml/qml/app.qml:9:1: module "QtQuick.Controls" version 1.3 is not in
stalled
```

mottosso:
Ah, yes. Pip only has 5.2 I think.

To install via pip, you can install directly from GitHub instead. But that's a separate issue. Merging this!

PR #200, opened 2016-05-26
removed if not plug.active

linez69:
Fixes Bug, if a plugin is set active=False and you manually activate it again, it won't get executed

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -277,8 +277,6 @@ class Controller(QtCore.QObject):
         }
 
         for plug, instance in pyblish.logic.Iterator(plugins, context):
-            if not plug.active:
-                continue
 
             state["nextOrder"] = plug.order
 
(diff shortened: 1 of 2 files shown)


mottosso:
Thanks @linez69!

Because it's quite a sensitive change, that might significantly change how plug-ins are run, I'll give this some time for full on testing before merging. You can help by keeping an eye out for things that act different, aside from what you're actually fixing.

tokejepsen:
What is the status of this PR? Looking to use the ```active``` property atm.

mottosso:
I haven't found any problems yet, but it would be great if you could test this feature with your setup @tokejepsen before we merge into master.

tokejepsen:
Seems to work great :smile: Will continue using this PR for some development, and see if there are any problems.

mottosso:
> Will continue using this PR for some development, and see if there are any problems.

That would be great!

I'm only holding this off because it's a subtle change and I'm trying not accidentally break anything. Once you feel it's stable, I'm happy to merge.

tokejepsen:
Half of development and no problems so far. Safe to merge for my sake.

PR #204, opened 2016-06-19
0.7.0

mottosso:
Stripped 2 out of 3 dependencies, only pyblish-base is required now!

Also added installation instructions to README. More to come.

**Before**

1. Install pyblish-rpc
2. Install pyblish-integration
3. Install pyblish-maya, or something like it.
4. Install pyblish-qml
5. Launch Pyblish QML in terminal
6. Start server within your host
7. Emit a `show` signal to Pyblish QML
8. Done.

**After**

1. Install pyblish-qml
2. Launch Pyblish QML in terminal
3. `pyblish_qml.show()`


tokejepsen:
Guessing this will the standalone as well?

mottosso:
> Guessing this will the standalone as well?

You mean replace [pyblish-standalone](https://github.com/pyblish/pyblish-standalone)? The roles haven't changed, just that you no longer will need to install RPC or Integration separately. Standalone should still be relevant in the same way.

tokejepsen:
> You mean replace pyblish-standalone?

Yeah, since we are getting rid on pyblish-maya etc., was thinking that pyblish-standalone would be merged as well?

mottosso:
pyblish-maya remains, it still contains plug-ins and helper functionality related to Maya. It's just the QML part being stripped, to make room for alternative GUIs, such as Lite.

tokejepsen:
Ahh, I see. Got confused from your **Before/After** :)

mottosso:
Cool. :) Does it make sense though? I'm looking for ways to make QML simpler, while at the same time open up for other GUIs. At the moment, pyblish-maya and other integrations are quite heavily dependent on QML.

tokejepsen:
It does make sense, and I think the simplicity is going to serve new people very well.

> At the moment, pyblish-maya and other integrations are quite heavily dependent on QML.

Guessing ideally you would just have some plugins for each integration?

mottosso:
Plug-ins that are common across integrations, and specific things like the menu item, yes. Anything specific to a particular host, basically.

The next challenge is how to get the menu item to come from something like pyblish-maya, but trigger pyblish-qml, pyblish-lite or any other GUI. Still thinking about that..

BigRoy:
How about something similar to `register_host ()` but for GUI. `register_gui ("QML", gui_method)`
Possibly passing it the method to instantiate it directly or passing it the command to run. `"import pyblish_qml; pyblish_qml.show ()"`
The latest registered gui could be the active one. Similarly one can list the available gui and enable another one again.

mottosso:
Thanks Roy, I'll give that a try.

mottosso:
> +6,400 −13,010

Nothing better than reducing the size of a project. <3

mottosso:
Ok, guys. It would be great if you could give these changes a try.

Have a look at the [README](https://github.com/mottosso/pyblish-qml/blob/972b846cfbeda724999743abd8c0853fe43af002/README.md) for how it's meant to be used.

mottosso:
@BigRoy done. Let me know what you think.

- https://github.com/pyblish/pyblish-maya/pull/57

Issue #209, opened 2016-08-04
Missing PyQt5 for pip install


tokejepsen:
**Description**

When installing with ```pip install pyblish-qml```, PyQt5 does not get installed.

**Expected results**

I would expect PyQt5 to be installed as a dependency, to work out-the-box.

**Short reproducible**

```
pip install pyblish-qml
python -m pyblish_qml --debug
```

tokejepsen:
I've got a ```setup.py``` working and tested, sort of:s

```pyblish_qml``` opens with nothing in it.
![capture](https://cloud.githubusercontent.com/assets/1860085/17409381/f5100150-5a66-11e6-85ea-cd030948c8e1.PNG)

And there seems to be a problem with "QtQuick.Controls":
```
(temp) C:\Users\admin\Documents\pipeline\launch>python -m pyblish_qml --debug
Starting Pyblish..
Starting in debug-mode
Looking for server..
Listening on 127.0.0.1:9090
No existing server found, creating..
Debug server created successfully.
Running mocked RPC server @ 127.0.0.1:6000
Listening on 127.0.0.1:6000Settings:

  WindowTitle = Pyblish (DEBUG)
  ContextLabel = World
  WindowSize = (430, 600)
Entering state: "hidden"
Entering state: "ready"
Entering state: "clean"
Entering state: "alive"
Awaited statemachine for 0.00 ms
Entering state: "visible"
Entering state: "initialising"
Spent 1204.00 ms creating the application
file:///C:/Users/admin/Documents/pipeline/miniconda/envs/temp/lib/site-packages/
pyblish_qml/qml/app.qml:9:1: module "QtQuick.Controls" version 1.3 is not instal
led
INFO - Collecting Negatron
INFO - Selecting instances..
INFO - Selecting: Peter01
INFO - Selecting: Richard05
INFO - Selecting: Steven11
INFO - Selecting: Piraya12
INFO - Selecting: Marcus
INFO - Selecting: Extra1
INFO - Selecting: DependencyInstance
INFO - Selecting: NoFamily
INFO - Selecting: Failure 1
WARNING - I'm good
INFO - Selecting: Failure 2
WARNING - I'm about to fail
INFO - Sleeping for 2 seconds..
INFO - Good morning
INFO - Reversing instances in the context..
INFO - Reversed!
INFO - Collecting Positron
Spent 2528.00 ms resetting
Made 19 requests during publish.
Spent 2825.00 ms resetting
[...]

PR #210, opened 2016-08-05
Adding PyQt5 as an install dependency

tokejepsen:
As discussed in #209. I only tested the latest PyQt5 5.7, so I don't know whether it'll work on older versions.

Now I'm guessing we should also inform the user somehow that installing ```pyblish_qml``` via. pip+git will need python 3.5, or is the current README section about requirements enough?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,7 +78,7 @@ setup(
     },
     install_requires=[
         "pyblish-base>=1.4",
-        "python-qt5>=0.1.8"
+        "PyQt5>=5.7"
     ],
     entry_points={},
 )
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,8 @@ setup(
         ]
     },
     install_requires=[
-        "pyblish-base>=1.4"
+        "pyblish-base>=1.4",
+        "python-qt5>=0.1.8"
     ],
     entry_points={},
 )


tokejepsen:
Guessing the CI is running on python 2 and not python 3?

tokejepsen:
It does become a problem when installing for python 2, as the install fails.

We could account for this by querying the python version; http://stackoverflow.com/questions/13924931/setup-py-restrict-the-allowable-version-of-the-python-interpreter ?

Also I was looking at using dependency links for installing ```python-qt5``` directly from github, but as @mottosso said its only for Windows, so we don't want to support it?

mottosso:
> It does become a problem when installing for python 2, as the install fails.

Ah. Did not expect that..

We'll have to find a way to make it work on versions of Python that don't have PyQt5 accessible via pip. Ideally without too much complication.

The goal is to make it simpler and/or more clear than how the instructions are currently in the README. 

> Also I was looking at using dependency links for installing python-qt5 directly from github, but as @mottosso said its only for Windows, so we don't want to support it?

To be fair, PyQt5 also has quite limited availability, so neither is ideal. If we can't make it simpler, it's better to leave it as is.

tokejepsen:
Alright, leave it as it is:)

Issue #213, opened 2017-01-12
Reuse running QML


mottosso:
### Goal

Improve robustness of pyblish-qml.

### Motivation

It's possible to launch multiple QML clients, which causes a host such as Maya to send commands sporadically, to whichever is the quickest to answer. They may end up picking every-other command, which is just broken behaviour.

### Implementation

When launching QML, first see if there already is a client running, and if there is, reuse it.

PR #214, opened 2017-01-12
Implement #213

mottosso:
#213


mottosso:
Usage.

**Terminal 1**

```bash
$ python -m pyblish_qml
```

**Terminal 2**

```bash
$ python -m pyblish_qml
```

End result, only one QML is only ever launched.

PR #217, opened 2017-02-09
Implement #216

mottosso:
#216 

pyblish-qml previously did not work with Pyblish 1.4.3 and above, due to the newly added `match` attribute of plug-ins.

This fixes that.


mottosso:
This also enables support for PyQt 5.7.1.

Issue #227, opened 2017-03-01
Alert user on completion


BigRoy:
# Problem

For long publishes it often happens one minimizes the Window, yet it is never brought to attention when the processing finishes (e.g. after Validation _with_ errors or after completing a valid publish)

# Solution

Let the icon in the taskbar blink, like many applications do to hint something has happened.

# Reference

https://gitter.im/pyblish/pyblish?at=58b6e89721d548df2c76a3ee
http://www.qtforum.org/article/12285/taskbar-blink-flash.html

mottosso:
Would you like a popup balloon to appear as well? Could also tell you whether publishing failed or succeeded.

![image](https://cloud.githubusercontent.com/assets/2152766/23466985/af768f24-fe94-11e6-85a5-208da497ceb2.png)

BigRoy:
It could definitely. I'm wondering whether a user would only want to see that whenever the Pyblish window was "inactive" at the time of finishing? (E.g. when validating it might be done quickly)

mottosso:
Sure, there could be a timer on it, say 10 seconds after leaving the window inactive.

BigRoy:
Having checked with the users in our studio they think it would be a great addition to have a pop-up to become visible. Maybe just try the pop-up _whenever it is inactive_ to begin with?

BigRoy:
Closing this to continue the balloon pop-up notification in another issue.

PR #228, opened 2017-03-01
Implement #227 

mottosso:
Alert the window on `controller.finished signal`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ build
 playground.py
 *.qmlc
 *.jsc
+.idea/
--- a/pyblish_qml/app.py
+++ b/pyblish_qml/app.py
@@ -85,6 +85,7 @@ class Application(QtGui.QGuiApplication):
 
         host = ipc.client.Proxy()
         controller = control.Controller(host)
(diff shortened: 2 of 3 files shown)


mottosso:
Thanks @BigRoy, should we try and get the balloon popup in there as well? Or better save that for a next PR?

BigRoy:
This is working great as a minor feature actually, because of its simplicity I assume this could be part of a patch version and separate out the notification. I'm expecting that to have some discussion along the way, regarding how long it stays, design, what's messaged in it and how. That would mean this minor feature will only be waiting longer than needed (and could stay off from merging until we've gone fully through the notification feature).

How about renaming #227 to _alert user on completion_ and make a new issue regarding _show notification pop-up upon completion_ to really have them separated?

mottosso:
> How about renaming #227 to alert user on completion and make a new issue regarding show notification pop-up upon completion to really have them separated?

Agreed, this is such a simple and small modification that it stands well on its own.

mottosso:
Version up please?

BigRoy:
I happened to have merged with this repo just now, and added something to .gitignore.

mottosso:
No problem.

PR #235, opened 2017-05-25
adding in support for window position update

Coyode:
fixing line endings


mottosso:
Would you happen to be using some additional software, aside from just git, to help you make pull requests? I see this one is coming from a new branch, `patch-2` whereas your previous PR is `patch-1` and I don't think you would have made those yourself.

What really should have happened is for you to make a second commit onto the same branch, that way the original PR would take this new commit into account and update itself.

In this case, it's not a problem, just something to think about.

Thanks!

PR #236, opened 2017-05-31
Set the context label and window title, if not already set externally.

tokejepsen:
Currently you can't set the context label or window title on startup of the applications.

### DIFF
--- a/pyblish_qml/host.py
+++ b/pyblish_qml/host.py
@@ -226,9 +226,10 @@ def _install_maya():
     app = QtWidgets.QApplication.instance()
     app.aboutToQuit.connect(_on_application_quit)
 
-    # Configure GUI
-    settings.ContextLabel = "Maya"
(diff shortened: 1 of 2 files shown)


pyblish_qml/host.py
```diff
@@ -226,9 +226,10 @@ def threaded_wrapper(func, *args, **kwargs):
     app = QtWidgets.QApplication.instance()
     app.aboutToQuit.connect(_on_application_quit)
 
-    # Configure GUI
-    settings.ContextLabel = "Maya"
-    settings.WindowTitle = "Pyblish (Maya)"
+    if settings.ContextLabel == "Context":
```
tokejepsen: We certainly can :) Done.

tokejepsen:
The travis-ci on this seems to have been stuck?

mottosso:
Looks like it's done, it's GitHub that hasn't updated. I think this looks good, welcome to merge!

PR #237, opened 2017-05-31
Support Pyblish Standalone

tokejepsen:
This is not working code, but I thought it would be best to start the discussion about whys and hows like this.

This is the current output, resulting in no interface:
```
$ python -m pyblish_standalone --register-gui pyblish_qml
Found gui: <module 'pyblish_qml' from 'c:\users\tje\documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-qml\pyblish_qml\__init__.pyc'>
Setting up Pyblish QML in Pyblish Standalone
$
```
I'm probably missing an ```app.exec_()```, but I don't know where to put it as putting it in ```_install_standalone``` makes the execution stop after ```Setting up Pyblish QML in Pyblish Standalone``` without an gui.

I'm also unsure of how the ```threaded_wrapper``` should work. Maybe it needs to be run in a separate thread?

### DIFF
--- a/pyblish_qml/host.py
+++ b/pyblish_qml/host.py
@@ -298,12 +298,12 @@ def _install_standalone():
     import pyblish_standalone
 
     def threaded_wrapper(func, *args, **kwargs):
-        return func(args, kwargs)
+        return func(*args, **kwargs)
 
     sys.stdout.write("Setting up Pyblish QML in Pyblish Standalone\n")
     register_dispatch_wrapper(threaded_wrapper)
 
-    app = QtWidgets.QApplication(sys.argv)
+    app = QtWidgets.QApplication.instance()
     app.aboutToQuit.connect(_on_application_quit)
 
     settings.ContextLabel = "Standalone"
--- a/pyblish_qml/host.py
+++ b/pyblish_qml/host.py
@@ -188,7 +188,8 @@ def install_host():
     for install in (_install_maya,
                     _install_houdini,
                     _install_nuke,
-                    _install_hiero):
+                    _install_hiero,
+                    _install_standalone):
         try:
             install()
         except ImportError:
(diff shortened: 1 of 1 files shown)


mottosso:
I'm not entirely sure there is a need for pyblish-standalone. pyblish-qml runs standalone as-is, and responds to plug-ins set via `PYBLISHPLUGINPATH`.

```python
# Standalone
$ set PYBLISHPLUGINPATH=c:\some\plugins
$ python -m pyblish_qml
```

What is it pyblish-standalone is doing aside from that?

tokejepsen:
> What is it pyblish-standalone is doing aside from that?

Having a common interface to the UI's. It emulating the DCC's ```gui.show()```, basically having boiler plate code for showing the UIs.

tokejepsen:
We could do a check to determine whether to use "python -m pyblish_qml" in pyblish_standalone?

mottosso:
> We could do a check to determine whether to use "python -m pyblish_qml" in pyblish_standalone?

Why not just call `python -m pyblish_qml` as-is?

tokejepsen:
> Why not just call python -m pyblish_qml as-is?

pyblish-standalone has some other features, like register host and setting the ```context.data["currentFile"]```. Can you do that with pybilsh-qml alone?

tokejepsen:
Since pyblish_qml is assuming an QApplication instance, I think pyblish_standalone should provide this: https://github.com/pyblish/pyblish-standalone/pull/15

mottosso:
I've been thinking about this, and due to how little pyblish-standalone is contributing was considering this.

```bash
$ set PYBLISHGUI=pyblish_qml
$ pyblish --gui
```

Where `pyblish` is the Pyblish executable you get from `pip install pyblish`, alternatively..

```bash
$ python -m pyblish --gui
```

That way, we would get `--data` for free.

```bash
$ python -m pyblish --data currentFile some_file.txt --gui
```

The implementation would be something along the lines of..

```python
def cli()
  gui = __import__(os.environ["PYBLISHGUI"])
  gui.show()
```

What do you think?

tokejepsen:
> What do you think?

I like it :)

> I've been thinking about this, and due to how little pyblish-standalone is contributing was considering this.

I'm curious about why register the gui in an environment variable?
If we are going down the cli route, wouldn't it make more sense to register the gui through the cli flags?

One thing that is current missing from the cli part of pyblish, is to register a host. I'm using this currently separate plugins coming from standalone. What do you think of adding ```--register-host``` flag?

tokejepsen:
> I've been thinking about this, and due to how little pyblish-standalone is contributing was considering this.

Or were you thinking that the ```PYBLISHGUI``` variable could contain multiple guis?

Picks up the first available gui from ```PYBLISHGUI```.
```
$ set PYBLISHGUI=pyblish_qml;pyblish_lite
$ pyblish --gui
```

tokejepsen:
This has been superseeded by https://github.com/pyblish/pyblish-base/pull/313

PR #247, opened 2017-07-28
Perspective cosmetics

tokejepsen:
There was a comma missing between the default family and the families.

**Before**
before

**After**
after

Single family is as before.
single

### DIFF
--- a/pyblish_qml/models.py
+++ b/pyblish_qml/models.py
@@ -384,7 +384,11 @@ class ItemModel(AbstractModel):
         item["hasCompatible"] = True
 
         # Visualised in Perspective
-        item["familiesConcatenated"] = instance["data"].get("family", "")
+        item["familiesConcatenated"] = ""
+        if instance["data"].get("family", ""):
+            item["familiesConcatenated"] += instance["data"].get("family", "")
+        if instance["data"].get("families", []):
+            item["familiesConcatenated"] += ", "
         item["familiesConcatenated"] += ", ".join(
             instance["data"].get("families", []))
 
--- a/pyblish_qml/models.py
+++ b/pyblish_qml/models.py
@@ -384,13 +384,9 @@ class ItemModel(AbstractModel):
         item["hasCompatible"] = True
 
         # Visualised in Perspective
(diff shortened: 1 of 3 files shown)


pyblish_qml/models.py
```diff
@@ -384,7 +384,11 @@ def add_instance(self, instance):
         item["hasCompatible"] = True
 
         # Visualised in Perspective
-        item["familiesConcatenated"] = instance["data"].get("family", "")
```
mottosso: A little too verbose I think. How about something like this?

```python
", ".join(
  family for family in
  [instance["data"].get("family")] +
  instance["data"].get("families", [])
  if family is not None
)
```

Could maybe make a function out of it.

```python
def families(instance):
  return list(
    family for family in
    [instance["data"].get("family")] +
    instance["data"].get("families", [])
    if family is not None
  )

", ".join(families(instance))
```

tokejepsen: Where do you think that function should live?

mottosso: It's for verbosity and readability only, so it could live in the same block. But I wouldn't go for a function, I'm mostly spitballing. So long as it reads better and is less verbose, it's ok.

tokejepsen: How about now?

pyblish_qml/models.py
```diff
@@ -384,9 +384,9 @@ def add_instance(self, instance):
...
-        item["familiesConcatenated"] += ", ".join(
-            instance["data"].get("families", []))
+        families = [instance["data"]["family"]]
```
mottosso: `family` is guaranteed to exists since 1.2.1 and normally I would say we should keep backwards compatibility across the entire 1.0 spectrum; but Pyblish QML has already disavowed the `repair` functionality since a few versions back, so in this case I'll accept this.

Could you add a note to the README about compatibility with 1.2.1 and above?

tokejepsen: Sure :) Done.

mottosso:
Ok, looks good! Thanks @tokejepsen

PR #254, opened 2017-08-17
Emit close signal

tokejepsen:
**Motivation**

This is to add the ability to execute code when the pyblish-qml window is closed.

**Implementation**

Maybe we should have a dedicated "pyblishQmlCloseForced" event emitted as well when closing forcibly?

Also I would like to write a test for the signal as well, but I find it hard to write test for pyblish-qml because I don't have access to Docker. Maybe we could have a ```run_tests.py``` for local testing?

### DIFF
--- a/pyblish_qml/app.py
+++ b/pyblish_qml/app.py
@@ -39,15 +39,14 @@ class Window(QtQuick.QQuickView):
             modifiers = self.parent.queryKeyboardModifiers()
             shift_pressed = QtCore.Qt.ShiftModifier & modifiers
             states = self.parent.controller.states
-            close_signal = "pyblishQmlClose"
 
             if shift_pressed:
                 print("Force quitted..")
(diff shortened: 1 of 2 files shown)


tokejepsen:
I also just realized that there are no documentation for the signals 😟

mottosso:
> I also just realized that there are no documentation for the signals 

It's here! https://api.pyblish.com/pages/Events.html

tokejepsen:
> It's here! https://api.pyblish.com/pages/Events.html

That's the signal emitted from pyblish-base. There does not seem to be any documentation for the signals emitted by pyblish-qml, like for example https://github.com/pyblish/pyblish-qml/blob/eefede9c224c7d9da6b6eb8e8b9b042694ea7275/pyblish_qml/control.py#L131

mottosso:
Looks good to me, if you can merge, go right ahead.

PR #256, opened 2017-08-18
Mistake in testing with Python 2.

tokejepsen:


mottosso:
Mistake? `nose` is the executable for Python 2, `nose3` or `nosetests3` is for Python 3

tokejepsen:
```nose``` does not seem to be available when ```pip install nose``` with Python 2?

**Comment by mottosso** [comment]
Ah, sorry. It should be `nosetests`!

Then indeed it is a mistake. The approach you've taken here is ok too, but in that case maybe can remove the separation between Python's, and just say `python -m` for both, and let the user choose which python to call?

**Comment by tokejepsen** [comment]
> The approach you've taken here is ok too, but in that case maybe can remove the separation between Python's, and just say python -m for both, and let the user choose which python to call?

You just this for both Python 2 and Python 3?
```
python -m nose --verbose --with-doctest --exe --exclude=vendor
```

mottosso:
> You just this for both Python 2 and Python 3?

Yeap

PR #258, opened 2017-08-22
Targets workflow.

tokejepsen:
**Motivation**

This PR is to facilitate the targets workflow that was introduced in https://github.com/pyblish/pyblish-base/releases/tag/1.5.0.

**Implementation**

The implementation for pyblish-qml is done so you can exclusively target plugins. For a demonstration you can execute:

```bash
$ python -m pyblish_qml --demo --targets studio
```

This will only load two plugins that are targeted directly for ```studio```.
To get all plugins you can target both ```default``` and ```studio```:

```bash
$ python -m pyblish_qml --demo --targets default studio
```

You can also specify the targets when showing pyblish-qml from a host:

```python
import pyblish_qml
pyblish_qml.show(targets=["studio"])
```

Because the matching is using ```pyblish.api.plugins_by_targets``` its important to get https://github.com/pyblish/pyblish-base/pull/321 merged first.

Lastly I'm not sure whether there is better way than passing the ```targets``` argument all the way through to the controller?

### DIFF
--- a/pyblish_qml/__init__.py
+++ b/pyblish_qml/__init__.py
@@ -5,9 +5,9 @@ from .version import (
 )
 
 
-def show(parent=None):
+def show(parent=None, targets=[]):
     from . import host
-    return host.show(parent)
+    return host.show(parent, targets)
 
 
 _state = {}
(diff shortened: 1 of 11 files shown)


tokejepsen:
Latest update facilitates registering new targets in the host.

```python
import pyblish_qml
import pyblish.api

pyblish.api.register_target("studio")
pyblish_qml.show()
```

tokejepsen:
This has been used for a week in production. Proceeding to merging.

PR #260, opened 2017-08-25
Fix #249 Mangled comment

### LINKED ISSUE #249: Mangled comment
The last two characters of a comment is left out from publishing.

To reproduce, enter anything into either the header and/or body of a comment and print the results of `context.data["comment"]`. You'll find that `Hello World!` becomes `Hello Worl`.

yamahigashi:
Hi Marcus,

This fixes #249 .

https://github.com/pyblish/pyblish-qml/blob/master/pyblish_qml/control.py#L603-L615
```python
    def on_commenting(self, summary, description):

        def update():
            comment = "\n".join( entry for entry in self.data["comment"] if entry)   # <= This and...

            context = self.host.cached_context
            context.data["comment"] = comment
            self.data.update({"comment": (summary, description)})  # <= This line must before `comment` declaration

```

### Cause
* There seems mismatch with updating host's context between updating its content of `self.data`.

### Result
* Host receive previous state of `self.data["comment"]`

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -601,6 +601,8 @@ class Controller(QtCore.QObject):
         """The user is entering a comment"""
 
         def update():
+            self.data.update({"comment": (summary, description)})
+
             comment = "\n".join(
                 entry for entry in self.data["comment"]
                 if entry
@@ -608,7 +610,6 @@ class Controller(QtCore.QObject):
 
             context = self.host.cached_context
             context.data["comment"] = comment
-            self.data.update({"comment": (summary, description)})
 
             # Notify subscribers of the comment
             self.host.update(key="comment", value=comment)
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -42,9 +42,7 @@ class Controller(QtCore.QObject):
     finished = QtCore.pyqtSignal()
     initialised = QtCore.pyqtSignal()
     commented = QtCore.pyqtSignal()
-    commenting = QtCore.pyqtSignal(str,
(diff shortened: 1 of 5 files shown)


mottosso:
Thanks @yamahigashi, giving this a try now.

mottosso:
I had a go with this myself a few days ago, and thought that the separation between Summary and Description is overly complex; both implementation wise and to the user.

I thought of making it look like this instead, which could simplify things for everyone.

image

What do you think?

yamahigashi:
>the separation between Summary and Description is overly complex; both implementation wise and to the user. I thought of making it look like this instead, which could simplify things for everyone.

I agree with this, it looks good to me.
I think that it is better to delegate handling of the comment body with actual plug-ins.

mottosso:
Ok, give me a minute and I'll update this PR.

mottosso:
Ok, let me know how that works for you!

yamahigashi:
It works fine, thx ; )
but I think `console.log(text);` should be removed in `CommentBox.qml`. https://github.com/yamahigashi/pyblish-qml/commit/bdc72728a977d0fb28121afb46dc6a431fd77be3#diff-6c927b464bd3f9ee8ea59b64b81681c3R73

mottosso:
> should be removed in CommentBox.qml

Oh! Yes, you are right. Debugging line. Thanks, removing that now.

mottosso:
For completeness, the mock plug-in added prints the comment during validation so you can test the code by running..

```bash
$ python -m pyblish_qml --demo
```

Merging shortly.

PR #265, opened 2017-09-13
Added support for publish and validate from within the host app

liorbenhorin:
Looking for the best way to call pyblish_qml's window actions, like publish, validate, etc
In pylish_lite, we can simply do:
```
window = pyblish_maya.show()
window.publish() #after validation...
```
After chatting [here ](http://forums.pyblish.com/t/call-pyblish-qml-actions/388 )with @mottosso, 
Here's what i came up with.
Usage:

```
import pyblish_qml.host

pyblish_qml.host.publish()
pyblish_qml.host.validate()
```

### DIFF
--- a/pyblish_qml/app.py
+++ b/pyblish_qml/app.py
@@ -70,6 +70,8 @@ class Application(QtGui.QGuiApplication):
     risen = QtCore.pyqtSignal()
     inFocused = QtCore.pyqtSignal()
     outFocused = QtCore.pyqtSignal()
+    published = QtCore.pyqtSignal()
+    validated = QtCore.pyqtSignal()
 
     def __init__(self, source, targets=[]):
         super(Application, self).__init__(sys.argv)
@@ -102,6 +104,8 @@ class Application(QtGui.QGuiApplication):
         self.risen.connect(self.rise)
         self.inFocused.connect(self.inFocus)
         self.outFocused.connect(self.outFocus)
+        self.published.connect(self.publish)
+        self.validated.connect(self.validate)
 
         window.setSource(QtCore.QUrl.fromLocalFile(source))
 
@@ -207,6 +211,14 @@ class Application(QtGui.QGuiApplication):
             previous_flags = self.window.flags()
             self.window.setFlags(previous_flags ^ QtCore.Qt.WindowStaysOnTopHint)
 
+    def publish(self):
+        """Fire up the publish sequence"""
+        self.controller.publish()
+
+    def validate(self):
+        """Fire up the validation sequance"""
+        self.controller.validate()
+
     def listen(self):
         """Listen on incoming messages from host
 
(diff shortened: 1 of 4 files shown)


mottosso:
Whop! This looks very good to me. :)

Are you testing this? @tokejepsen would you be able to give this a test?

tokejepsen:
This works :) And quite nicely as well!

@liorbenhorin could I ask your use-case for this? Why not use ```pyblish.api```?

liorbenhorin:
Hi @tokejepsen!

My goal was to start the publish_qml and that the sequence will start automatically, without the user having to press anything.
Only if the test fails, then the GUI will remain open to see what went wrong.

If this can be achieved from `pyblish.api` i would love to know how :)

tokejepsen:
> If this can be achieved from pyblish.api i would love to know how :)

You wouldn't be able to get the GUI as ```pyblish.api``` is all non-GUI, but you would be able to produce a report to your user of any validation failures.
I've been experimenting with this myself; https://github.com/tokejepsen/pyblish-grill-environment/blob/7230a561cfbbe419aeb8eebaad90282c62dc1512/environment/PYTHONPATH/grill_tools/nuke/pyblish_init.py#L123
Although when there is a validation failure, there user would have to go into the GUI anyways to maybe fix it with an action. Your solution here might be better :)

tokejepsen:
I'm guessing your instances are already pre-enabled?

mottosso:
I think @tokejepsen is referring to..

```python
from pyblish import util
context = util.publish()
```

Which then has an embedded report.

```python
for result in context["results"]:
  if not result["success"]:
    print("%s failed" % result["plugin"])

    for record in result["records"]:
      print(record.msg)
```

- [Reference](https://learn.pyblish.com/chapters/15-report-i.html)

liorbenhorin:
@tokejepsen 
For now the usage is very basic, a single instance, say a character, being 'pushed' as a new master, so yeah, it's active (by enabled you mean active? I'm new here!)
I will make sure to look into your method soon!

@mottosso 
Yup, I'm aware of the non-gui method, and I'm using it for batch operations!

tokejepsen:
> (by enabled you mean active? I'm new here!)

That's right :) You might run into problems when people wan't to pick'n'choose what to publish, but in your simple case this would work well.

liorbenhorin:
@tokejepsen got it.
Yeah Iv'e got more complexed scenarios but they don't need to run automatically.

tokejepsen:
Think we can merge this :)
@mottosso do you wanna do the honors?

mottosso:
Happy to, but can someone up the version first? Then we can release this. :)

liorbenhorin:
@mottosso done

mottosso:
Thanks @liorbenhorin!

tokejepsen:
Missing a release for this :)

mottosso:
Indeed, setting one up now.

Issue #218, opened 2017-02-09
Inactive optional plugin won't run


derrickb:
This could be related to a similar issue I filed on pyblish-lite: https://github.com/pyblish/pyblish-lite/issues/90

I have an optional Validator, the first in the list, that I want to be off by default as it could take a long time to complete for larger scenes.

I have these properties on the plugin:
```
optional = True
active = False
```

If I attempt to run it either by itself or with the other plugins disabled, it doesn't get applied:
![image](https://cloud.githubusercontent.com/assets/1793769/22807414/afd017ce-eedb-11e6-8620-38a1418df05f.png)
![image](https://cloud.githubusercontent.com/assets/1793769/22807418/b3f7e55c-eedb-11e6-8f2f-fc9c6134068d.png)

However if I change "active" to "True", then it does work:
![image](https://cloud.githubusercontent.com/assets/1793769/22807426/be8019f4-eedb-11e6-8ba7-1ddae89c4fdd.png)

mottosso:
Thanks for the detailed report @derrickb. 

This looks to be some problem with the defaulting of the visual toggle in the GUI. `active` is the value the toggle is manipulating, and `optional` is what enables the user to manipulate it. By the looks of it, the toggle isn't getting the initial value of `active` as it is being loaded into the GUI.

Having a look at this as soon as possible.

mottosso:
Which version are you on?

```python
>>> import pyblish_qml
>>> print(pyblish_qml.__version__)
0.7.1
```

derrickb:
0.7.1

mottosso:
This has now been fixed in 0.8.1

For completeness and future self, the issue was that the Pyblish Iterator is using the `active` flag as a means of determining whether or not to process a plug-in, but QML did not update this variable. Since QML runs remotely, it proved tricky to try and modify this member and so instead I implemented a slimmed down version of the iterator specifically for QML.

For now, it's an ok solution, but it does defeat the initial purpose of a common iterator. Namely that if behaviour changes I'll now need to update both base and qml separately.

mottosso:
This should also re-enable the use of `instance.data["publish"] = False` for defaulting the toggled status of instances.

PR #266, opened 2017-10-21
Implement pyblish-base/#250

BigRoy:
Implement the same functionality for QML as for base regarding: https://github.com/pyblish/pyblish-base/issues/250

This should work with PR: https://github.com/pyblish/pyblish-base/pull/325

---

Note: Somehow when toggling off an instance of family `a` it will still show the ContextPlugins that are filtered to family a, but it will not run them. How to visually hide them I have no clue, but aside from cosmetics it is working.

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -692,12 +692,12 @@ class Controller(QtCore.QObject):
                                                                   plugin)
                     plugin.compatibleInstances = list(i.id for i in instances)
                 else:
-             
 
                     # When filtering to families at least a single instance 
                     # with that family must be available for ContextPlugin
-                    if ("*" in plugin.families or 
-                          pyblish.logic.instances_by_plugin(context, plugin)):
+                    has_wildcard = "*" in plugin.families
+                    has_one = pyblish.logic.instances_by_plugin(context, plugin) != []
+                    if has_wildcard or has_one:
                         plugin.compatibleInstances = [context.id]
 
             self.data["models"]["item"].reorder(context)
@@ -1052,14 +1052,12 @@ def iterator(plugins, context):
                 yield plugin, instance
 
         else:
-        
-     
             # When filtering to families at least a single instance with
             # that family must be active in the current publish
-            if "*" not in plugin.families:
-                if not any(instance.data.get("publish") is not False 
-                           for instance in instances):
-                    continue
-                    
-        
+            no_wildcard = "*" not in plugin.families
+            no_active_instance = not any(inst.data.get("publish") for inst in instances)
+
+            if no_wildcard and no_active_instance:
+                continue
+
             yield plugin, None
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -692,7 +692,13 @@ class Controller(QtCore.QObject):
                                                                   plugin)
                     plugin.compatibleInstances = list(i.id for i in instances)
                 else:
-                    plugin.compatibleInstances = [context.id]
+             
+
+                    # When filtering to families at least a single instance 
+                    # with that family must be available for ContextPlugin
+                    if ("*" in plugin.families or 
+                          pyblish.logic.instances_by_plugin(context, plugin)):
+                        plugin.compatibleInstances = [context.id]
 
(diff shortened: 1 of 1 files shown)


tokejepsen:
Will this also work with the matching algorithms?

BigRoy:
Since it triggers the `instance_by_plugins` logic I'd assume it does, but I haven't tested it.

BigRoy:
@tokejepsen are you able to test this? :) We have this live in production here right now, but are not using `targets` workflows and neither to we use the other matching algorithms.

tokejepsen:
Seems to work just fine here :)

pyblish_qml/control.py
```diff
@@ -1046,4 +1052,14 @@ def iterator(plugins, context):
                 yield plugin, instance
 
         else:
+        
+     
+            # When filtering to families at least a single instance with
+            # that family must be active in the current publish
+            if "*" not in plugin.families:
+                if not any(instance.data.get("publish") is not False
```
mottosso: I can't make sense of this line.

If not <any of these> is not <Not True>, then loop through the below. Have a look and see if you can't make this more intuitive

BigRoy: The check is this explicit to check for `False` because I believe the Pyblish logic does the same. As such I couldn't just do `if not instance.data.get("publish")` because then it would also consider `None` whereas the previous Pyblish logic wouldn't.

The logic is that _at least one instance must have its publish set to True_. So if there's no "active" one it should `continue`.

The check whether it's true is: `instance.data.get("publish") is not False`

This could also work:
```python
def _is_active(instance):
    """Whether instance is currently set active to publish"""
    return instance.data.get("publish") is not False 

if not any(_is_active(instance) for instance in instances):
    continue
```

mottosso:
Ok, I took a stab at simplifying the logic. It's still complex, and I haven't tested it. Let me know what you think.

pyblish_qml/control.py
```diff
@@ -1052,14 +1052,12 @@ def iterator(plugins, context):
...
             # When filtering to families at least a single instance with
             # that family must be active in the current publish
-            if "*" not in plugin.families:
-                if not any(instance.data.get("publish") is not False 
-                           for instance in instances):
-                    continue
-                    
-        
+            no_wildcard = "*" not in plugin.families
+            no_active_instance = not any(inst.data.get("publish") for inst in instances)
```
BigRoy: @mottosso note that this is now a different check than what `logic.py` does in `pyblish-base` because there it explicitly checks for `is not False`. This could potentially start behaving different.

mottosso: Yes. Do try this locally first, then we'll need to update Lite and Base to follow suit. It's a backwards incompatible change, so odds are we'll put it behind a `PYBLISH_SOMETHING_SOMETHING` environment flag to be explicitly enabled.

mottosso:
The `iterator()` you just removed is from [here](https://github.com/pyblish/pyblish-qml/blame/master/pyblish_qml/control.py#L1058), which leads to this: https://github.com/pyblish/pyblish-qml/issues/218#issuecomment-280087728

Make sure we don't break that.

BigRoy:
> Make sure we don't break that.

Like this? :smile:

mottosso:
> Interesting that the previous also passed the tests, that should've been caught by a test.

This project is lacking a good test suite, which is also why it's a little risky. But if you could give this a go yourself and keep an eye out for oddities, I think we should be fine.

BigRoy:
Ok, thanks!

Any idea how to solve this one: https://github.com/pyblish/pyblish-qml/pull/266/commits/6a20765ea8f91d4aec9c58dbde2341913ad44726#diff-ac156cba73ac1ef385b357bbfc21d258R698 ?

In [pyblish base I made this backwards compatible by implementing a toggle in `os.environ`](https://github.com/pyblish/pyblish-base/pull/325/commits/55a96202f80fd5b05ffd49dec0cb7582c16dcb57) but I'm not sure how to propagate that to pyblish QML since it runs in its own process. The QML process does not inherit that value from the parent process so I can't just seem to check against that variable.

PR #267, opened 2017-11-05
fix instance Warning state behavior in GUI

davidlatwe:
Found that instance's warning state only take effect from the last plugin.
* If warning was not emitted by the last plugin, then it will be override by ~~error or *(it should be)*~~ success.

	

* If warning was emitted by the last plugin, instance's error state will be override.

	

---

Thought this should not be a normal ~~instance item~~ log level behavior, I added `persistWarning` into item-model to keep the warning state for instance item, and reset warning state if that instance got error.

---

[Simple test case I used](https://gist.github.com/davidlatwe/37766b311e708a2df988c15852c8fe09)

### DIFF
--- a/pyblish_qml/models.py
+++ b/pyblish_qml/models.py
@@ -466,7 +466,7 @@ class ItemModel(AbstractModel):
                 item.succeeded = True
                 item.amountPassed += 1
 
-            if not (item.hasWarning and not item.hasError):
+            if item.hasError:
                 item.hasWarning = False
 
             item.duration += result["duration"]
--- a/pyblish_qml/models.py
+++ b/pyblish_qml/models.py
@@ -453,7 +453,7 @@ class ItemModel(AbstractModel):
             item.isProcessing = False
             item.currentProgress = 1
             item.processed = True
-            item.hasWarning |= any([
+            item.hasWarning = item.hasWarning or any([
                 record["levelno"] == logging.WARNING
                 for record in result["records"]
             ])
--- a/pyblish_qml/models.py
+++ b/pyblish_qml/models.py
@@ -454,7 +454,7 @@ class ItemModel(AbstractModel):
             item.isProcessing = False
             item.currentProgress = 1
             item.processed = True
-            item.hasWarning = item.persistWarning or any([
+            item.hasWarning |= any([
                 record["levelno"] == logging.WARNING
                 for record in result["records"]
             ])
--- a/pyblish_qml/models.py
+++ b/pyblish_qml/models.py
@@ -466,9 +466,6 @@ class ItemModel(AbstractModel):
                 item.succeeded = True
                 item.amountPassed += 1
 
-            if item.hasError:
-                item.hasWarning = False
-
             item.duration += result["duration"]
             item.finishedAt = time.time()
 
--- a/pyblish_qml/models.py
+++ b/pyblish_qml/models.py
@@ -17,7 +17,6 @@ defaults = {
         "familiesConcatenated": "",
         "isToggled": True,
         "hasWarning": False,
-        "persistWarning": False,
         "hasError": False,
         "actionHasError": False,
         "actionPending": True,
@@ -467,11 +466,8 @@ class ItemModel(AbstractModel):
                 item.succeeded = True
                 item.amountPassed += 1
 
-            if item.hasWarning and not item.hasError:
-                item.persistWarning = True
-            else:
+            if not (item.hasWarning and not item.hasError):
                 item.hasWarning = False
-                item.persistWarning = False
 
             item.duration += result["duration"]
             item.finishedAt = time.time()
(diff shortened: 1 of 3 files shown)


davidlatwe:
Since QML doesn't count the warning amount, so keep it count them as passed.

pyblish_qml/models.py
```diff
@@ -453,7 +454,7 @@ def update_with_result(self, result):
             item.isProcessing = False
             item.currentProgress = 1
             item.processed = True
-            item.hasWarning = any([
+            item.hasWarning = item.persistWarning or any([
```
mottosso: Could this be:

```python
item.hasWarning = item.hasWarning or any([...
```

That way, it'll be set once, and then re-used. That way we'd save on introducing an additional variable too.

mottosso:
Well spotted!

pyblish_qml/models.py
davidlatwe: Oh, yes ! That line can change to
```python
item.hasWarning |= any([...
```

davidlatwe: Done, is this what you were saying ?

davidlatwe:
Can I also fix my last PR's pep8 issue ?

pyblish_qml/models.py
davidlatwe: Oh, wait, yes, I might not need that `persistWarning `

davidlatwe: Done ! Awesome !

mottosso: Actually, I would prefer `item.hasWarning = item.hasWarning or ...`

Because, I'm not sure if you know, `|=` does not mean "or equals". It means "bitwise or equals" and I'm really rusty on my bitwise math. I'm not sure you knew `True |= False == True` which it does in this case and should, but to me that's black magic and I will be looking at that line sometime in the future wondering "What does that do?".

It's not as pretty as what you've got, but I think it will save someone time in the future. Does that make sense?

pyblish_qml/models.py
```diff
@@ -466,6 +466,9 @@ def update_with_result(self, result):
                 item.succeeded = True
                 item.amountPassed += 1
 
+            if not (item.hasWarning and not item.hasError):
```
mottosso: What does this do?

davidlatwe: Oh, it reset the warning, if this instance got error from any plugin.
Without this reset, warning will override the error state.

davidlatwe: Since this
```python
item.hasWarning |= any([ ...
```
lock the warning state, so need that

```python
if not (item.hasWarning and not item.hasError):
```
to take warning back for exposing the error state.

davidlatwe: Oh, sorry, that should be able simplify to
```python
if item.hasError:
    item.hasWarning = False
```

pyblish_qml/models.py
davidlatwe: > I'm not sure if you know, |= does not mean "or equals".

I am not aware that !
Sure, of course.

davidlatwe: It's in fix 4 now.

pyblish_qml/models.py
```diff
@@ -466,6 +466,9 @@ def update_with_result(self, result):
                 item.succeeded = True
                 item.amountPassed += 1
 
+            if item.hasError:
+                item.hasWarning = False
```
mottosso: Ah, does it show yellow even if there's an error? In that case it sounds like an error in the GUI code. Shouldn't need to alter the warning when there's an error.

Having a look..

Aah, yes. If you shuffle [this statement](https://github.com/pyblish/pyblish-qml/blob/7e86c446c8d196527bd7a5d47a0a5f8d5683287b/pyblish_qml/qml/List.qml#L35), to query the error first, then the `hasError` will take prevalence and it won't matter whether it has a warning.

Give it a try.

davidlatwe: > Aah, yes. If you shuffle this statement, to query the error first, then the hasError will take prevalence and it won't matter whether it has a warning.

Yes ! It did !

davidlatwe: Wow, much clearer now.

davidlatwe: it's in fix 5 now.

mottosso:
Excellent, this looks good!

In the future, it'd be better to put a commit message that describes what the change does, rather than a series of versions. Think of each commit as a self-contained whole, something you might one day want to revert, or cherry-pick onto a different branch. And consider what it will look like in midst other commits, [here](https://github.com/pyblish/pyblish-qml/commits/master).

For now, this is fine. Just a minor tweak for future PRs.

mottosso:
Thanks @davidlatwe!

davidlatwe:
> In the future, it'd be better to put a commit message that describes what the change does
Sure, will do !

Issue #264, opened 2017-09-13
SystemError: error return without exception set


tokejepsen:
**Issue**

We quite often get threading issue with pyblish-qml and Nuke with the following message:

```bash
Traceback (most recent call last):
File "C:/Program Files/Nuke10.5v1/plugins\nuke\executeInMain.py", line 19, in executeInMainThreadWithResult
r = _nuke.RunInMainThread.result(id)
SystemError: error return without exception set
```

Threading issues are always tough to replicate. but users seem to experience them in bulks. This error makes pyblish-qml window open and then immediately close, which makes the users try it again and again in sequence.
Seems like the issue can resolve itself by waiting and/or trying multiple times.

For these reasons maybe it could be something to do with a pyblish-qml instance that does not get closed/shutdown properly?

mottosso:
Without being able to reproduce it, I'm not sure what I can do. :( If you can manage to reproduce the error outside of Nuke then I'd be able to narrow it down.

Alternatively, you could narrow down the scope of the program inside of Nuke, by for example launching a simpler program with threading and try to catch it with as little code as possible.

For example..

```python
import threading

def my_program():
  print("Hello world")

thread = threading.Thread(target=my_program)
thread.start()
```

And then try adding things to the program until you experience the bug.

tokejepsen:
Maybe other people will experience this eventually, although I suspect it might be something to do with the collectors I'm using Nuke. Did try to run through the collection * 100, and didn't experience any problems with ```pyblish.util.collect()```.

This only error I experienced when investigating was this:

```python
import time

import pyblish_qml


server = pyblish_qml.show()
time.sleep(5)
server.stop()
```
```
# Result: Using Python @ 'C:\Users\tje\AppData\Local\Continuum\Miniconda2\envs\tgbvfx-environment\python.exe'
Using PyQt5 @ 'None'
Targets: default
Success. QML server available as pyblish_qml.api.current_server()
Starting pyblish-qml
Done, don't forget to call `show()`
Entering state: "hidden"
Entering state: "ready"
Entering state: "clean"
Entering state: "alive"
Settings: 
  WindowTitle = pipeline_test / sq001 / sh0020 / plateprep
  HeartbeatInterval = 60
  WindowPosition = [0, 0]
  WindowSize = [1280, 1344]
  ContextLabel = Nuke
Exception in thread Thread-23:
Traceback (most recent call last):
[...]

PR #269, opened 2017-11-09
Modal mode

tokejepsen:
This PR implements a modal mode and a way to switch between modal and non-modal.

The motivation for a modal mode is because of threading issues in Nuke reported in #264.

**Usage**

Non-modal is the default when using ```pyblish_qml.show()```.

To invoke modal mode you can either explicitly call ```pyblish_qml.show(modal=True)``` or set the environment variable ```os.environ["PYBLISH_QML_MODAL"] = "True"```.

### DIFF
--- a/pyblish_qml/__init__.py
+++ b/pyblish_qml/__init__.py
@@ -5,8 +5,11 @@ from .version import (
 )
 
 
-def show(parent=None, targets=[], modal=None):
+def show(parent=None, targets=None, modal=None):
     from . import host
+
+    if targets is None:
+        targets = []
     return host.show(parent, targets, modal)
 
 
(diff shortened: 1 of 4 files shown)


pyblish_qml/__init__.py
```diff
@@ -5,9 +5,9 @@
 )
 
 
-def show(parent=None, targets=[]):
+def show(parent=None, targets=[], modal=None):
```
mottosso: Iiiih while you're at it, could you remove the `[]` in targets?

Pass `None` and convert it to a list within function instead. Lists in signature arguments are mutable, which means if you modify the list, the next call the the function (without `targets=`) will behave differently. Very difficult to debug!

mottosso:
This looks good. Just that one nitpick, and then please version up and release.

PR #271, opened 2017-11-30
Bug fix for standalone.

tokejepsen:
Will need some help explaining why this fixes the issue 😕


tokejepsen:
When creating the modal mode, we removed the server call to ```self.listen()```, which is why we need to add it when not running as a child (or inside a host).

mottosso:
`listen()` is what enables QML to chat with whatever host launches it; it'll need to be called at some point regardless of how it's run. I can't tell exactly what's happening here; so long as it doesn't cause trouble for running things in non-standalone mode, then this looks fine.

tokejepsen:
Non-standalone checked, and working fine :)

PR #277, opened 2017-12-14
Check icon

mottosso:
Along the theme of section colors, I wanted to see what it'd look like to update the checkbox icon with something other than the check, as check normally means "all good", which isn't the case when things error.

untitled project

Not sure it's an improvement.. Welcome to tinker.

As I was writing this, I discovered a minor bug in the previously merged PR; namely that the section icon went too dark when unexpanded and had no status.

untitled project

### DIFF
--- a/pyblish_qml/qml/Pyblish/CheckBox.qml
+++ b/pyblish_qml/qml/Pyblish/CheckBox.qml
@@ -10,6 +10,7 @@ MouseArea {
 
     property bool active: true
     property bool checked: false
+    property alias icon: check.name
 
     property var statuses: {
         "default": "white",
--- a/pyblish_qml/qml/Pyblish/ListItems/SectionItem.qml
+++ b/pyblish_qml/qml/Pyblish/ListItems/SectionItem.qml
@@ -15,7 +15,7 @@ Item {
     property string text
 
     property var statuses: {
-        "default": "#222",
+        "default": "#ddd",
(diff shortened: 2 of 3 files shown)


tokejepsen:
That is an interesting idea!

Could we maybe remove the box around the exclamation icon?
And how about the deprecation icon for warnings?

BigRoy:
Not sure if it's better - I think it's more clutter. 

How about *after* processing that the checkboxes turn "fully colored" instead of showing an icon - they would just be red, orange or green. And can't be toggled.

tokejepsen:
> How about after processing that the checkboxes turn "fully colored" instead of showing an icon

+1

tokejepsen:
Here are the solid colors in action:

dec 18 2017 11_09 am

Its in a PR for @mottosso fork; https://github.com/mottosso/pyblish-qml/pull/3

mottosso:
Oh wow, that looks great. Merging that.

tokejepsen:
I have another PR which just cleans the added code and versions up, ready for a merge; https://github.com/mottosso/pyblish-qml/pull/4

mottosso:
Thanks @tokejepsen, happy to merge this. Any last comments?

tokejepsen:
I'm good for a merge :) @BigRoy ?

BigRoy:
Looks awesome visually. Great work.

PR #284, opened 2018-03-21
Fix #283

### LINKED ISSUE #283: Unchecked instances visually unclear
### Issue

Not sure where this got introduced, but in the latest version of Pyblish QML unchecked instances are visually hard to read because the checkbox remains a fully solid checked state.

worse

In this image the `cameraDefault` instance is actually disabled, it's clicked to be deactivated. 

The only hint that visualizes it is disabled is the darker greyed out text. 

Previously it would also clear the solid contents of the checkbox and keep only the border, which visually made much more sense.

### Proposal

Remove the solid checked color when deactivated, like:

better

---

Could be from [this change](https://github.com/pyblish/pyblish-qml/pull/277/commits/63c55d3384644e3d2733dcc583ae6308410c8d95) which was [shown here](https://github.com/pyblish/pyblish-qml/pull/277#issuecomment-352383187).

BigRoy:
This fixes #283 

Though, on toggling an instance off for a very brief moment the "checked" icon is visible. Not sure if that's a problem, but it's because that one fades over a brief time: 
```
NumberAnimation {
                duration: 100
            }
```

Current result: https://i.gyazo.com/9ecf5c88571b859c2953772c4c00bd33.mp4

### DIFF
--- a/pyblish_qml/qml/Pyblish/CheckBox.qml
+++ b/pyblish_qml/qml/Pyblish/CheckBox.qml
@@ -59,12 +59,14 @@ MouseArea {
         anchors.centerIn: parent
 
         color: {
-          if (status == "success")
-            return Qt.darker(check.color, 1.5)
-          if (status == "warning")
-            return Qt.darker(check.color, 1.5)
-          if (status == "error")
-            return Qt.darker(check.color, 1.5)
+          if (checked) {
+              if (status == "success")
+                return Qt.darker(check.color, 1.5)
+              if (status == "warning")
+                return Qt.darker(check.color, 1.5)
+              if (status == "error")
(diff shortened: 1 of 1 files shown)


davidlatwe:
image

How about we make the border color be dimmed white `Qt.darker("white", 1.5)` as well when it's unchecked ?

davidlatwe:
Oh, ignore me if keeping the border color is intended. :P

mottosso:
Missed this! Happy for this to get merged once you guys are happy. 👍

BigRoy:
Let's merge this! Can't see a merge button on my phone, so not sure if I have the privileges to do so. :)

Issue #285, opened 2018-04-02
Section collapsing indicator color not working properly with action `Category`


davidlatwe:
## Problem
If I add `pyblish.api.Category` object in to `plugin.actions` list, the color of section collapsing indicator will not working as expected (Seems like plugin status got override to *default*).

## Test Code

Please try comment out `pyblish.api.Category("Fix"),` in each plugin's `actions` and see the change of section indicator `+`

```python
import pyblish.api


class DoSomeFix(pyblish.api.Action):
    label = "Make Some"
    on = "failed"

    def process(self, context, plugin):
        pass


class CollectSome(pyblish.api.ContextPlugin):

    order = pyblish.api.CollectorOrder - 0.2
    label = "Collect Some"
    actions = [
        #pyblish.api.Category("Fix"),
        DoSomeFix,
    ]

    def process(self, context):
        context.create_instance("something")


class CollectFailed(pyblish.api.ContextPlugin):

    order = pyblish.api.CollectorOrder - 0.2
    label = "Collect Failed"
    actions = [
        pyblish.api.Category("Fix"),
        DoSomeFix,
    ]

    def process(self, context):
        raise RuntimeError("Failed.")


class ValidateSome(pyblish.api.InstancePlugin):

    order = pyblish.api.ValidatorOrder - 0.2
    label = "Validate Some"
    actions = [
        pyblish.api.Category("Fix"),
        DoSomeFix,
    ]

    def process(self, instance):
        raise RuntimeError("Failed.")

```

mottosso:
Thanks @davidlatwe, would you like to have a look at resolving this?

davidlatwe:
Sure, will send a PR if I solved this.

davidlatwe:
Solved with #286
Closing this one now:)
Thanks!

PR #286, opened 2018-04-02
Changed actionsIconVisible loop breaking for section indicator updating

davidlatwe:
After adding hide-able section function, updating `actionsIconVisible` state
may skip section update if the action menu exists.

So instead of `continue if True` which skipping the rest of the code,
changed to `if not True`.

This will solve #285 .

### DIFF
--- a/pyblish_qml/models.py
+++ b/pyblish_qml/models.py
@@ -499,9 +499,7 @@ class ItemModel(AbstractModel):
             item.duration += result["duration"]
             item.finishedAt = time.time()
 
-            if item.itemType == "plugin":
-                if item.actionsIconVisible:
-                    continue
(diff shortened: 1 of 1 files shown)


mottosso:
I stared at these two gifs for some time wondering what the difference was. :D Is it that the `+` icon of the section doesn't turn red when an action fails?

image

davidlatwe:
Ah, yes, should have put some note about that 😨 
The `+` icon wont change the color if the plugin which has action icon visible, done with warning or error.

mottosso:
Ah ok, that makes sense. :) Thanks for this. Could you also increment the version to 1.7.3? (hopping over 1.7.2, which has been mistakenly released without incrementing).

davidlatwe:
Done ! :D

Issue #291, opened 2018-07-10
"optional" instance data key not being interpreted


darkvertex:
In pyblish-lite it's easy enough to set `.data['optional'] = False` to make an instance mandatory, but it seems to not be handled any differently in `pyblish-qml`.

Is this a known issue? I couldn't find an open ticket about it.

Any hint where in the code the problem might be? (Unless it's an easy/obvious fix to do on your end.)

I looked a bit and maybe it's not here, but I thought it was weird that in:
https://github.com/pyblish/pyblish-qml/blob/master/pyblish_qml/qml/List.qml#L24
```
        active: object.optional
```
It's strange to me that the listitem being "active" is driven by the object's "optional" attribute, no? (Forgive me if I'm misinterpreting as my QML chops are low.)

tokejepsen:
Are you talking about a plugin or instance?

Plugins use ```optional``` and ```active``` data members to determine if the plugin is optional (togglable) and active (toggled by default or not)

Instances are always optional. The data member ```publish``` determines whether an instance is toggled by default or not.

darkvertex:
I'm talking about an Instance.

> Instances are always optional.
In pyblish-lite they can be non-toggleable/non-optional if `instance.data['optional'] = False`. (Usually this is combined with `instance.data['publish'] = True` to make sure they appear toggled ON.)

mottosso:
I'm also quite surprised that functionality existed in Lite, but it sounds useful. Happy to see this implemented in QML.

darkvertex:
> I'm also quite surprised that functionality existed in Lite

![pyblishlite_mandatory_instances](https://user-images.githubusercontent.com/941331/42574696-088bdfa8-84ed-11e8-844d-1ec787851f6f.gif)

I found it by pure curiosity when making my first few Pyblish publishers (for which I initially used "lite") and when it worked as well as it did, I just assumed it was a feature! haha. :laughing:

darkvertex:
Implemented by PR https://github.com/pyblish/pyblish-qml/pull/292 which has been merged and is part of the [1.7.4 release.](https://github.com/pyblish/pyblish-qml/releases/tag/1.7.4) (woohoo!)

PR #292, opened 2018-07-11
Add support for mandatory instances

darkvertex:
`pyblish-lite` implements this by omitting the checkbox when an Instance has `.data['optional'] = False`, effectively making an instance "mandatory" (in whatever state it was added.)

With this PR, `pyblish-qml` implements mandatory isntances by greying out the checkbox, just like it's already doing for the "Context" entry in the top left. Clicking on them shows a "Cannot toggle." status prompt.

To show what this looks like, here's a dummy Collector I used:
```python
class Collector(pyblish.api.ContextPlugin):
    def process(self, context):
        instance = context.create_instance('farmer', family='human')
        instance.data['optional'] = False
        instance.data['publish'] = True
        for i in range(3):
            instance = context.create_instance('potato%s' % i, family='vegetable')
            instance.data['optional'] = True
            instance.data['publish'] = False
```

And here's what this looks like in action:

pyblishqml_mandatory_instances

and for comparison, the same Collector without the PR's changes, in action:

pyblishqml_mandatory_instances_not_working

and the same Collector again, with the latest official release of `pyblish-lite`:

pyblishlite_mandatory_instances

If this was to be merged, it would fix issue #291.

### DIFF
--- a/pyblish_qml/ipc/formatting.py
+++ b/pyblish_qml/ipc/formatting.py
@@ -133,6 +133,9 @@ def format_data(data):
         "publish",
         "comment",
 
+        # Allows an instance to be non-optional (ie, mandatory)
+        "optional",
+
         # Provided by service.py
         "host",
         "port",
--- a/pyblish_qml/version.py
+++ b/pyblish_qml/version.py
@@ -1,7 +1,7 @@
 
 VERSION_MAJOR = 1
 VERSION_MINOR = 7
-VERSION_PATCH = 3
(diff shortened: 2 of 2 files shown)


darkvertex:
@mottosso Just bumped VERSION_PATCH in `version.py`

(Is that the only place for the version number info? Let me know if I have to edit elsewhere.)

mottosso:
That's exactly right, sorry I should have said!

This all looks good to me, thanks again!

PR #293, opened 2018-08-15
Containerizing subprocess GUI to host widget

davidlatwe:
### It's real child window in host now !

pyblish_qml

Require some help on testing different platform, currently only tested on Windows 7 and 10 in Maya 2016, 2018.

(Thanks to this [gist](https://gist.github.com/mmolero/c3db18390aa71bb6b3cdcb9822483032#file-cefpython_pyside-py-L60-L73))

---

This PR changed the relationship between host and client a lot, the implementation and the variable naming might not the best, any suggestion are welcome :)

### DIFF
--- a/pyblish_qml/host.py
+++ b/pyblish_qml/host.py
@@ -41,7 +41,7 @@ def current_server():
     return _state.get("currentServer")
 
 
-def install(modal):
+def install():
     """Perform first time install
 
     Attributes:
@@ -55,7 +55,7 @@ def install(modal):
         uninstall()
 
     install_callbacks()
-    install_host(modal)
+    install_host()
 
     _state["installed"] = True
 
@@ -77,6 +77,22 @@ def _is_headless():
     )
 
 
+def acquire_host_main_window():
+
+    app = QtWidgets.QApplication.instance()
+
+    # Get top window in host
+    _window = app.activeWindow()
+    while True:
+        parent_window = _window.parent()
+        if parent_window:
+            _window = parent_window
+        else:
+            break
+
+    _state["vesselParent"] = _window
+
+
 def show(parent=None, targets=[], modal=None):
     """Attempt to show GUI
 
@@ -91,18 +107,21 @@ def show(parent=None, targets=[], modal=None):
 
     # Automatically install if not already installed.
     if not _state.get("installed"):
-        install(modal)
+        install()
+        acquire_host_main_window()
 
     # Show existing GUI
     if _state.get("currentServer"):
         server = _state["currentServer"]
-        proxy = ipc.server.Proxy(server)
+        server.modal = modal
 
         try:
+            proxy = _state["currentProxy"]
+            proxy.update(server)
             proxy.show(settings.to_dict())
             return server
 
-        except IOError:
+        except (IOError, KeyError):
             # The running instance has already been closed.
             _state.pop("currentServer")
 
@@ -140,6 +159,7 @@ def show(parent=None, targets=[], modal=None):
 
     # Store reference to server for future calls
     _state["currentServer"] = server
+    _state["currentProxy"] = proxy
 
     print("Success. QML server available as "
           "pyblish_qml.api.current_server()")
@@ -153,12 +173,13 @@ def publish():
     # get existing GUI
     if _state.get("currentServer"):
         server = _state["currentServer"]
-        proxy = ipc.server.Proxy(server)
 
         try:
+            proxy = _state["currentProxy"]
+            proxy.update(server)
             proxy.publish()
 
-        except IOError:
+        except (IOError, KeyError):
             # The running instance has already been closed.
             _state.pop("currentServer")
 
@@ -167,12 +188,13 @@ def validate():
     # get existing GUI
     if _state.get("currentServer"):
         server = _state["currentServer"]
-        proxy = ipc.server.Proxy(server)
 
         try:
+            proxy = _state["currentProxy"]
+            proxy.update(server)
             proxy.validate()
 
-        except IOError:
+        except (IOError, KeyError):
             # The running instance has already been closed.
             _state.pop("currentServer")
 
@@ -227,7 +249,7 @@ def register_pyqt5(path):
     _state["pyqt5"] = path
 
 
-def install_host(modal):
+def install_host():
     """Install required components into supported hosts
 
     An unsupported host will still run, but may encounter issues,
@@ -242,7 +264,7 @@ def install_host(modal):
                     _install_hiero,
                     _install_nukestudio):
         try:
-            install(modal)
+            install()
         except ImportError:
             pass
         else:
@@ -264,89 +286,7 @@ def _on_application_quit():
         pass
 
 
-def _connect_host_event(app):
-    """Connect some event from host to QML
-
-    Host will connect following event to QML:
-    QEvent.Show                -> rise QML
-    QEvent.Hide                -> hide QML
-    QEvent.WindowActivate      -> set QML on top
-    QEvent.WindowDeactivate    -> remove QML on top
-    """
-
-    class HostEventFilter(QtWidgets.QWidget):
-
-        eventList = [
-            QtCore.QEvent.Show,
-            QtCore.QEvent.Hide,
-            QtCore.QEvent.WindowActivate,
-            QtCore.QEvent.WindowDeactivate
-        ]
-
-        def getServer(self):
-            server = None
-            try:
-                server = _state["currentServer"]
-            except KeyError:
-                # No server started
-                pass
-            return server
-
-        def eventFilter(self, widget, event):
-            if event.type() in self.eventList:
-                server = self.getServer()
-                if not server:
-                    return False
-                else:
-                    proxy = ipc.server.Proxy(server)
-
-                if event.type() == QtCore.QEvent.Show:
-                    try:
-                        proxy.rise()
-                        return True
-                    except IOError:
-                        # The running instance has already been closed.
-                        _state.pop("currentServer")
-                if event.type() == QtCore.QEvent.Hide:
-                    try:
-                        proxy.hide()
-                        return True
-                    except IOError:
-                        # The running instance has already been closed.
-                        _state.pop("currentServer")
-                if event.type() == QtCore.QEvent.WindowActivate:
-                    try:
-                        proxy.inFocus()
-                        return True
-                    except IOError:
-                        # The running instance has already been closed.
-                        _state.pop("currentServer")
-                if event.type() == QtCore.QEvent.WindowDeactivate:
-                    try:
-                        proxy.outFocus()
-                        return True
-                    except IOError:
-                        # The running instance has already been closed.
-                        _state.pop("currentServer")
-            return False
-
-    # Get top window in host
-    app_top_window = app.activeWindow()
-    while True:
-        parent_window = app_top_window.parent()
-        if parent_window:
-            app_top_window = parent_window
-        else:
-            break
-    # install event filter
-    try:
-        host_event_filter = HostEventFilter(app_top_window)
-        app_top_window.installEventFilter(host_event_filter)
-    except Exception:
-        pass
-
-
-def _install_maya(modal):
+def _install_maya():
     """Helper function to Autodesk Maya support"""
     from maya import utils
 
@@ -355,15 +295,13 @@ def _install_maya(modal):
             func, *args, **kwargs)
 
     sys.stdout.write("Setting up Pyblish QML in Maya\n")
-    if not modal:
-        register_dispatch_wrapper(threaded_wrapper)
+    register_dispatch_wrapper(threaded_wrapper)
 
     app = QtWidgets.QApplication.instance()
 
     if not _is_headless():
         # mayapy would have a QtGui.QGuiApplication
         app.aboutToQuit.connect(_on_application_quit)
-        _connect_host_event(app)
 
     if settings.ContextLabel == settings.ContextLabelDefault:
         settings.ContextLabel = "Maya"
@@ -371,7 +309,7 @@ def _install_maya(modal):
         settings.WindowTitle = "Pyblish (Maya)"
 
 
-def _install_houdini(modal):
+def _install_houdini():
     """Helper function to SideFx Houdini support"""
     import hdefereval
 
@@ -380,12 +318,10 @@ def _install_houdini(modal):
             func, *args, **kwargs)
 
     sys.stdout.write("Setting up Pyblish QML in Houdini\n")
-    if not modal:
-        register_dispatch_wrapper(threaded_wrapper)
+    register_dispatch_wrapper(threaded_wrapper)
 
     app = QtWidgets.QApplication.instance()
     app.aboutToQuit.connect(_on_application_quit)
-    _connect_host_event(app)
 
     if settings.ContextLabel == settings.ContextLabelDefault:
         settings.ContextLabel = "Houdini"
@@ -393,7 +329,7 @@ def _install_houdini(modal):
         settings.WindowTitle = "Pyblish (Houdini)"
 
 
-def _install_nuke(modal):
+def _install_nuke():
     """Helper function to The Foundry Nuke support"""
     import nuke
 
@@ -410,12 +346,10 @@ def _install_nuke(modal):
             func, args, kwargs)
 
     sys.stdout.write("Setting up Pyblish QML in Nuke\n")
-    if not modal:
-        register_dispatch_wrapper(threaded_wrapper)
+    register_dispatch_wrapper(threaded_wrapper)
 
     app = QtWidgets.QApplication.instance()
     app.aboutToQuit.connect(_on_application_quit)
-    _connect_host_event(app)
 
     if settings.ContextLabel == settings.ContextLabelDefault:
         settings.ContextLabel = "Nuke"
@@ -423,7 +357,7 @@ def _install_nuke(modal):
         settings.WindowTitle = "Pyblish (Nuke)"
 
 
-def _install_nukeassist(modal):
+def _install_nukeassist():
     """Helper function to The Foundry NukeAssist support"""
     import nuke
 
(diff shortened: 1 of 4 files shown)


pyblish_qml/host.py
```diff
@@ -77,6 +77,22 @@ def _is_headless():
     )
 
 
+def acquire_host_main_window():
+
+    app = QtWidgets.QApplication.instance()
+
+    # Get top window in host
+    _window = app.activeWindow()
+    while True:
```
mottosso: I can see what you're trying to do, but there's an edge-case here that's a little dangerous. Which is that when the active window is a window that isn't parented to Maya (like a third-party, custom window, or poorly designed native window) then this would return that window rather than the main Maya window.

```python
bad_window = QtWidgets.QDialog()
bad_window.show()
# Now this window can be active, and wouldn't return any parent because it has none.
```

pyblish_qml/ipc/client.py
```diff
@@ -76,6 +78,12 @@ def emit(self, signal, **kwargs):
     def update(self, key, value):
         self._dispatch("update", args=[key, value])
 
+    def _self_destruct(self):
+        """Auto quit exec if parent process failed
+        """
+        self._kill = threading.Timer(30, lambda: os._exit(0))
```
mottosso: How come it's in a timer?

pyblish_qml/ipc/client.py
```diff
@@ -105,6 +113,10 @@ def _listen():
                     elif response.get("header") == "pyblish-qml:popen.parent":
                         self.channels["parent"].put(line)
 
+                    elif response.get("header") == "pyblish-qml:server.pulse":
+                        self._kill.cancel()  # reset timer
```
mottosso: Ahaa. :) Nice one.

Is `30` seconds or milli-seconds? Ms might be a little short, in case a pulse is delayed for some reason.

mottosso:
Wuuuw! I've tried doing this so many times but never succeeded. Nice work. :)

Giving this a spin soon.

pyblish_qml/host.py
davidlatwe: This one is a bit dangerous indeed !
The alternative I can think of is a piece of code from Avalon
```python
parent = {
    widget.objectName(): widget
    for widget in QtWidgets.QApplication.topLevelWidgets()
}[main_window_name]  # `main_window_name` require further input
```
Will this one better ?

pyblish_qml/ipc/client.py
davidlatwe: Opss, it 30 second.

The server will send pulse every 15 second, and the client here will reset every 30 second.
I'll append doc string about this to `_self_destruct` in next commit :P

pyblish_qml/host.py
mottosso: Yes, that will work.

`main_window_name` is constant in Maya 2013-2019+ and I don't expect it to change, I would put just `"MayaWindow"`

davidlatwe: Cool !  I'll try to find the main window name of Nuke and other expensive friends,
or do you have that list ?

davidlatwe: This seems does not have a quick solution for other DCC Tools :/
Update Maya first !

mottosso:
Just testing this out, and ran into an issue which reminded me of one of the benefits of keeping the window  separate; it keeps bugs out of Maya as well.



In this case, Pyblish QML froze without an error message, taking Maya down with it.

I can't actually get this to run at all, it reaches the above point with this in the console.

```bash
Setting up Pyblish QML in Maya
Using Python @ 'C:\Python36\python.exe'
Using PyQt5 @ 'None'
Targets: default
Success. QML server available as pyblish_qml.api.current_server()
Starting pyblish-qml
Done, don't forget to call `show()`
Entering state: "hidden"
Entering state: "alive"
Entering state: "clean"
Entering state: "ready"
  HiddenSections = ['Collect']
  WindowPosition = [100, 100]
  HeartbeatInterval = 60
  ContextLabel = Maya
  WindowSize = [430, 600]
  WindowTitle = Pyblish (Maya)
Settings: 
```

When I kill the associated Python process, Maya becomes responsive again, with this output.

```bash
# # Exception in thread Thread-3:
# Traceback (most recent call last):
#   File "C:\Program Files\Autodesk\Maya2015\bin\python27.zip\threading.py", line 551, in __bootstrap_inner
#     self.run()
#   File "C:\Program Files\Autodesk\Maya2015\bin\python27.zip\threading.py", line 504, in run
#     self.__target(*self.__args, **self.__kwargs)
#   File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-qml\pyblish_qml\ipc\server.py", line 326, in _listen
#     self.popen.stdin.write(data + b"\n")
# IOError: [Errno 22] Invalid argument
```

Which is valid; this is Maya attempting to communicate with a dead process. The not-so-valid thing is that it freezes.

@davidlatwe any idea what is going on?

mottosso:
That was Maya 2015.

In 2018, the problem is a little different.



The error message is different, and Maya doesn't freeze.

```python
Setting up Pyblish QML in Maya
Using Python @ 'C:\Python36\python.exe'
Using PyQt5 @ 'None'
Targets: default
Success. QML server available as pyblish_qml.api.current_server()
  File "C:\Python36\lib\runpy.py", line 193, in _run_module_as_main
  File "C:\Python36\lib\runpy.py", line 85, in _run_code
    "__main__", mod_spec)
  File "C:\Python36\lib\urllib\request.py", line 88, in <module>
    from urllib.request import urlopen
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-qml\pyblish_qml\vendor\jsonschema\compat.py", line 18, in <module>
    from .compat import str_types, MutableMapping, urlsplit
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-qml\pyblish_qml\vendor\jsonschema\_utils.py", line 6, in <module>
    from . import _utils
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-qml\pyblish_qml\vendor\jsonschema\exceptions.py", line 6, in <module>
    from .exceptions import (
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-qml\pyblish_qml\vendor\jsonschema\__init__.py", line 12, in <module>
    from ..vendor import six, jsonschema
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-qml\pyblish_qml\ipc\schema.py", line 16, in <module>
    from . import schema
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-qml\pyblish_qml\ipc\formatting.py", line 7, in <module>
    from . import mocking, formatting
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-qml\pyblish_qml\ipc\service.py", line 13, in <module>
    from . import client, server, service
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-qml\pyblish_qml\ipc\__init__.py", line 1, in <module>
    from . import util, compat, control, settings, ipc
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-qml\pyblish_qml\app.py", line 15, in <module>
    from . import app
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-qml\pyblish_qml\__main__.py", line 6, in <module>
    exec(code, run_globals)
ModuleNotFoundError: No module named 'http.client'; 'http' is not a package
    __import__(name)
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-base\pyblish\vendor\six.py", line 82, in _import_module
    return _import_module(self.mod)
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-base\pyblish\vendor\six.py", line 115, in _resolve
    result = self._resolve()
  File "C:\Users\marcus\Dropbox\dev\alienfx\github\pyblish\pyblish-base\pyblish\vendor\six.py", line 92, in __get__
    from six.moves import http_client
  File "C:\Program Files\Autodesk\Maya2018\plug-ins\MASH\scripts\googleapiclient\http.py", line 23, in <module>
    import http.client
```

davidlatwe:
Edit: Add the 2018 error workaround permalink

davidlatwe:
Looks like subprocess can spew by changing these two line in `ipc/server.py` [L297](https://github.com/pyblish/pyblish-qml/blob/a1446ecedb0d4e23214e1f26143a028c4e2d48a7/pyblish_qml/ipc/server.py#L297) and [L333](https://github.com/pyblish/pyblish-qml/blob/a1446ecedb0d4e23214e1f26143a028c4e2d48a7/pyblish_qml/ipc/server.py#L333)

from `sys.stdout.write(line)` to simple `print(line)`.

I am not sure where `sys.stdout` leads to, but my console did get more output from `print`

davidlatwe:
Ah ! I think I found the way to raise all error (the error silenced in slots)!

based on this [stack overflow Ans](https://stackoverflow.com/questions/18740884/preventing-pyqt-to-silence-exceptions-occurring-in-slots)

Will commit this for you to test after I clean up the code.

davidlatwe:
Just in case misunderstanding, the solution has tested on my end by increment `winId`  intentionally (so it can not find the container window) to produce error when `Application.show` is processing, which called by `pyqtSignal`.

Here's the error feedback
```
Starting pyblish-qml
Done, don't forget to call `show()`
AttributeError: 'NoneType' object has no attribute 'setWidth'
    vessel.setWidth(client_settings["WindowSize"][0])
  File "C:\Users\David\Dropbox\github\AVALON\pyblish-qml\pyblish_qml\app.py", line 168, in show
    func(*args)
  File "C:\Users\David\Dropbox\github\AVALON\pyblish-qml\pyblish_qml\app.py", line 37, in wrapper
Entering state: "alive"
Entering state: "clean"
Entering state: "ready"
```

This wasn't able to print before, could give enough hint for us to debug !

davidlatwe:
Hi @mottosso 
I got a fresh Maya 2018.0 installed and have a few test, added those new commits, please have a look if you are available :)

There's one other thing confused me, during my test in Maya 2018.0, I found that if I launch Maya from the `C:\Program Files\Autodesk\Maya2018\bin\`, which is default current working dir path in `.link`, in such condition, `from PyQt5 import ...` will raise `DLL load failed: The specified procedure could not be found` error in subprocess where it's about to init the QML app.

And If I set the `cwd` to any other path to `subprocess.Popen()`, then the error is gone.

Wonder has anyone happened before ?

mottosso:
> I got a fresh Maya 2018.0 installed and have a few test

Thanks for the diligence! This works for me on Windows 10 in Maya 2015 and 2018; I have since updated to 2018 Update 4 so wasn't able to test whether it worked with Update 0.

Some notes.

1. When killing Maya with Pyblish QML open, it automatically dies with it. Nice!
1. When the Pyblish QML Python process is killed the Maya parent window remains open. I killed it by hand this time around, but since Python 3 an exception automatically kills the process, so this could happen from time to time for the end-user. Could we have it close alongside the Python process? <br> 
1. When Maya is busy, Pyblish QML is non-responsive. This is unfortunate and somewhat of a dealbreaker to me personally; publishing is often a long-running and blocking process for me. Is there any way to make QML remain responsive?<br> 

Finally, after a few runs, I don't know what happened, but it has stopped working now. :S Maya freezes on launch, similar to the previous problem I had.

What's more interesting, although I can't move or resize windows via the titlebar, I'm using a utility application for moving an resizing applications via a hotkey (AltDrag) and it reveals what Maya is able to respond to resizing just fine, which suggest that it isn't frozen so much as just blocking input.

untitled project

Not sure what to make of that.

> And If I set the cwd to any other path to subprocess.Popen(), then the error is gone.

Ah, that is a good find. We should have that set per default for the process launching QML; anything but the Maya directory as we can never be sure what Maya chooses to put there to screw up things like PyQt5. I've added this to the PR.

davidlatwe:
Thanks for the feedback @mottosso  :D
Appreciated !

> Could we have it close alongside the Python process? 

I have added, with a little refactoring on my previous commits of this PR.

> When Maya is busy, Pyblish QML is non-responsive. This is unfortunate and somewhat of a dealbreaker to me personally; publishing is often a long-running and blocking process for me. Is there any way to make QML remain responsive?

I do think of some approach, but am afraid they all decrease the stability : (

#### launch freeze

> Finally, after a few runs, I don't know what happened, but it has stopped working now. :S Maya freezes on launch, similar to the previous problem I had.

I made a log file output to tail all messages processed in `server.listen` and `proxy.dispatch`, and I found the frozen point is right on main thread executing function **`emit`**. Did not freeze on every `show()` time, but the chance is not small.

This happens to be a similar error that I encountered.

#### log clearing freeze

Clear out Maya script editor log after QML is on ready state, and then press *reset* or *publish* or *validate*. In this case, the frozen point is when main thread executing function **`stats`**. (This exists before this PR)

*launch freeze* and *log clearing freeze*, both did not happen on Maya 2018 during my tests, and to those Maya version below 2018 whom freezes, after I manually kill the subprocess, their function returned, and the server wrote the last response to `popen.stdin` on both cases.

#### Experimental workaround for launch freeze

In the last commit, I found that use `default_wrapper` to execute function `emit` seems help the situation, the *launch freeze* is gone in Maya 2015, not 100% sure, at least did not happen in 20+ `pyblish_qml.show()` calling. 

I think the main problem is Maya's `utils.executeInMainThreadWithResult`, since no more frozen happened in Maya 2018.

### This PR may be in a tar pit :/

The issues comes after this implementation is a bit hard to resolve, at least for me it is, at this very moment.
I am gonna evaluate what changes may take as improvement for the master branch, and drop the implementation of subprocess GUI containerizing.
What do you think ? @mottosso 

Maya versions, tough crowd : (

mottosso:
That's allright, I'm thinking about this too. :)

I wonder if there's a way to spawn a child window and use it as a canvas Windows to draw into from a separate process? Where the drawing is done by Windows, and screen coordinates are provided by the window. That way, drawing would be independent of whether Maya is responsive, and coordinates would not change when it's frozen anyway.

In some ways, I feel as though what we have currently (a "fake" child) is what we would have ended up with if we had implemented this "real" child first. The fake child isn't half-bad, have you had any issues with it that prompted you to try a "real" window approach?

davidlatwe:
> That way, drawing would be independent of whether Maya is responsive, and coordinates would not change when it's frozen anyway.

What I had in mind was similar to this (if I understand correctly) :P
But then I remind that Maya's main window sometimes can be minimized if idle event exists in the long process like rendering, possible not a good patch :/

Now I am wondering if the container window can be spawned by a thread which started by the host (Maya). Currently the container window (the "real" child) is spawned in Maya's main thread, if I can push it to other thread, maybe the issue could be solved ?

> The fake child isn't half-bad, have you had any issues with it that prompted you to try a "real" window approach?

The issue is just that QML window only able to coexists with host main window, and not able to normally coexists with other child window (tricky). But this is only a minor issue compare to my TODO list in hand (or everyone, I believe  :P).

The thing really prompt me is the gist I found to convert `window id` to `int` in Python, which is the key element in this approach, which was the thing that I did not know then choosing "fake" child approach as alternative. Just wondering if the UX can be improved a bit thats all : )

mottosso:
> if I can push it to other thread, maybe the issue could be solved ?

No this won't work, because drawing is done via the main thread, and if it's stuck then it doesn't matter where it draws.

Having said that, you may be on to something. There is a [`processEvents`](http://doc.qt.io/qt-5/qcoreapplication.html#processEvents) which says..

> You can call this function occasionally when your program is busy performing a long operation (e.g. copying a file).

Though I wonder whether this applies to when the main thread is actually CPU bound. Worth a try? A quick google reveals that Qt is at least *able* to host windows in multiple threads.

- https://stackoverflow.com/questions/9777911/how-do-i-create-a-window-in-different-qt-threads

> Just wondering if the UX can be improved a bit thats all : )

For sure, I have tried doing exactly this for a number of projects before and never gotten this far. Regardless of whether it makes it into QML, it's a great addition to the mental toolbox in other projects so big thanks for sharing!

davidlatwe:
> Worth a try?

Definitely !  Thanks for the input !
Will get back with more result :D

davidlatwe:
Okay, I've got something able to share.

Host containerized GUI still able to get updated on each idle events which exists in between two jobs' execution end-start, but yes, very laggy. After some tests and google, turns out, you really can't do anything to improve GUI update speed as long as main thread is busy.

> that Qt is at least able to host windows in multiple threads.
> https://stackoverflow.com/questions/9777911

Well, from my learning from that QA, we can create things where ever we want, but you have to move it to main thread if you want it to be rendered. No luck !

So the only way that I can think of, is to detach it from host and parent back to the app which runs in subprocess, then parent to host again when the host main thread is free.

This approach (workaround) probably is a hack.

In my tests, it's not stable on Windows 7 in every version of Maya, but seems stable on Windows 10 with Maya 2017+ (the host seems required PySide2 or PyQt5 installed).

So I made this implementation optionable, using flag `foster` or environ var `PYBLISH_QML_FOSTER` to enable this experimental feature.
```python
pyblish_qml.show(foster=True)
```
Everything will stays the same if the feature is not enabled.

What do you think ? Is this experimental feature able to be accepted ?

davidlatwe:
Add one demo gif : )

qml_foster

mottosso:
Great research, @davidlatwe, I especially like this:

> detach it from host and parent back to the app which runs in subprocess, then parent to host again when the host main thread is free.

Ninja stuff! :D The disappearing/reappearing of the window when pressing the "Publish" button is a little unfortunate, but whether it's a problem in practice (e.g. makes the button scary to press) is another story, one we'd need to test it in practice for.

> Everything will stays the same if the feature is not enabled... What do you think ? Is this experimental feature able to be accepted?

Yes, that is excellent. This will let the feature be vetted on several setups before becoming a new default. I suspect it won't behave or even work on a system other than Windows, and expect we'll need to implement it again; possibly for each (major) window manager on Linux and at least once for OSX (possibly more, if OS versions differ much).

If you've incremented the version, feel free to merge!

davidlatwe:
Hooray ! Thanks : )
I have bumped the version, thanks again @mottosso .

mottosso:
Released, and ready for public testing.

```bash
$ pip install pyblish-qml==1.8.0
```

https://github.com/pyblish/pyblish-qml/releases/tag/1.8.0

Thanks @davidlatwe!

tokejepsen:
Nice work @davidlatwe !

Issue #294, opened 2018-08-28
Release 1.8.0 appears to be critically broken in Linux


darkvertex:
My studio's current Linux flavour is `CentOS Linux release 7.4.1708 (Core)`.

pyblish_qml release 1.8.0 (with the changes from PR #293) appears to be broken on my environment, throwing a KeyError before ever appearing. Release 1.7.4 does not have a problem.

Here's an environment with PyQt 5.6, `pyblish_base==1.6.1` and `pyblish_qml==1.8.0`:
```
> python
Python 2.7.13 (default, Apr  3 2017, 18:26:10)
[GCC 4.8.2 20140120 (Red Hat 4.8.2-15)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyblish_qml;pyblish_qml.show()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/softwareLocal/rez_python_envs/shared/2.7/0xb44bdba184a58db9/lib/python2.7/site-packages/pyblish_qml/__init__.py", line 13, in show
    return host.show(parent, targets, modal, foster)
  File "/softwareLocal/rez_python_envs/shared/2.7/0xb44bdba184a58db9/lib/python2.7/site-packages/pyblish_qml/host.py", line 125, in show
    install(modal, foster)
  File "/softwareLocal/rez_python_envs/shared/2.7/0xb44bdba184a58db9/lib/python2.7/site-packages/pyblish_qml/host.py", line 72, in install
    install_event_filter()
  File "/softwareLocal/rez_python_envs/shared/2.7/0xb44bdba184a58db9/lib/python2.7/site-packages/pyblish_qml/host.py", line 281, in install_event_filter
    main_window = _state["vesselParent"]
KeyError: 'vesselParent'
>>>
```
and no UI ever appears.

Here is the same, except using `pyblish_qml==1.7.4` instead of 1.8.0:
```
> python
Python 2.7.13 (default, Apr  3 2017, 18:26:10)
[GCC 4.8.2 20140120 (Red Hat 4.8.2-15)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyblish_qml;pyblish_qml.show()
Using Python @ '/rdo/software/rez/packages/python/2.7.13/platform-linux/bin/python'
Using PyQt5 @ 'None'
Targets: default
Success. QML server available as pyblish_qml.api.current_server()
[...]

PR #295, opened 2018-08-28
Hot fix for v1.8.0

davidlatwe:
Fix event filter install outside supported hosts.

### DIFF
--- a/pyblish_qml/host.py
+++ b/pyblish_qml/host.py
@@ -271,7 +271,6 @@ def install_host(use_threaded_wrapper):
         except ImportError:
             pass
         else:
-            install_event_filter()
             break
 
 
@@ -279,9 +278,8 @@ def install_event_filter():
 
     main_window = _state.get("vesselParent")
     if main_window is None:
-        sys.stdout.write("Main window not found, event filter did not "
-                         "installed.\n")
-        return
+        raise Exception("Main window not found, event filter did not "
+                        "install. This is a bug.")
 
     try:
         host_event_filter = HostEventFilter(main_window)
@@ -415,6 +413,8 @@ def _install_maya(use_threaded_wrapper):
             for widget in QtWidgets.QApplication.topLevelWidgets()
         }["MayaWindow"]
 
+        install_event_filter()
+
     _set_host_label("Maya")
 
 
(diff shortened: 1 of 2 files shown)


darkvertex:
I confirm this fixes issue #294 on my end and the UI appears as it normally would.

I haven't yet played with the foster mode. (If it's broken in Linux I'll make a separate ticket.)

davidlatwe:
The error was because `install_event_filter()` get called under unsupported host, and `vesselParent` is the main window of the host, which will only be set in supported hosts.

Thanks @darkvertex  :)

darkvertex:
I see! The explanation is appreciated.

When you speak of "supported hosts" do you mean DCCs like Maya, Houdini, etc? Does "Python" (standalone Python) count as a supported host in this context? I mean, I know Pyblish works with standalone Python, but I guess the hosts concept is for running from inside another software, correct?

davidlatwe:
> When you speak of "supported hosts" do you mean DCCs like Maya, Houdini, etc?

Yes, but to be more specific, I mean the host with GUI :)

davidlatwe:
I added one more commit, for the cases like running inside `mayapy`.
Sorry for the trouble guys.

pyblish_qml/host.py
```diff
@@ -277,7 +277,11 @@ def install_host(use_threaded_wrapper):
 
 def install_event_filter():
 
-    main_window = _state["vesselParent"]
+    main_window = _state.get("vesselParent")
+    if main_window is None:
+        sys.stdout.write("Main window not found, event filter did not "
```
darkvertex: I understand this changes nothing about the code logic and is kind of insignificant, but if you want to be more grammatically correct about "did not installed", this can be better written as one of two ways:
- `Main window not found, event filter did not get installed.`
- `Main window not found, event filter did not install.`

mottosso:
Oh! I didn't notice there was an event filter on the main window involved; that is a problem.

An event filter on the main window is a major performance bottleneck, as it would funnel every single Qt event - from user input to draw calls - through Python. There are hundreds of thousands of them, sometimes per second, and Python just isn't up for it. It affects things such as just clicking in the viewport; you may notice a slight 10-50 ms delay with an event filter installed.

The worst thing about it is that there is no way for the end-user to know why things have suddenly gotten slow. No way to debug.

It looks like it's only used for the foster child feature, could we make sure this is only installed when that feature is used? Preferably we'd find another way of accomplishing whatever it is doing without using an event filter.

mottosso:
Oh no! Thanks for confirming, I'll file an issue for it so we can have a look at resolving it.

In that case, this PR looks good.

davidlatwe:
Thanks @mottosso  :)
Then I guess it's okay to bump the version for merge ?

mottosso:
Yes please. :)

davidlatwe:
Ok, it's done.

PR #297, opened 2018-09-01
Implemented eventFilter removal and improved Foster stability

davidlatwe:
In this PR, should have:
* eventFilter auto remove after quitting under any conditions
    - No matter `foster` is on or not, usual window close or force quit, should remove evenFilter in the end.
* Foster mode should work under any versions of Maya
    - Behavior may differ by the Qt version on host, in my observation, host which running Qt4 could not render QML window properly while window setParent back to subprocess (*window detach-attach*), which meant for avoiding main thread busy. So in this PR, in order to guaranty the visual feedback, those Qt4 host will not be able to avoid main thread busy (lag better then blank).
    - For test propose, you can turn *detach-attach* on or off by setting environment variable `PYBLISH_QML_FOSTER_NINJA=True` or `False` (`Yes/No 1/0`)
* In any condition, window should able to ignore closeEvent while processing, unless force quitting
   - This was broken in my previous PR. :(
* Host main window should blink alert on validation/publish complete when in foster mode
   - In foster mode, the alert can only be shown on host main window.
* Fake child approach's OS limitation removed
   - Was only enabled on Windows, limitation is gone now.

Here are some Foster mode running screen gif:

This is Maya 2018.0 on Windows

qml_foster_windows

This is Maya 2015 on Windows

qml_foster_2015

As you can see, it's really laggy.

This is Maya 2017 update 5 on CentOS 7 (Run in Hyper-V)

qml_foster_linux

Foster mode works on CentOS !
One thing to be noticed is that, while in my first test on CentOS, I installed Maya 2017 and encountered an error says:
`FosterVessel(QtWidgets.QDialog) has no attribute 'winId'`
This is a bug, the error was gone after I installed the Maya 2017 update.

Now, I wouldn't say the Foster mode is 100% safe to use, but should at least stable as Fake child approach.

Any feedback are welcome :)
Thanks !

### DIFF
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -85,7 +85,7 @@ class Proxy(object):
     def hide(self):
         """Hide the GUI"""
         self._dispatch("hide")
-        self.vessel.hide()
+        return self.vessel.hide()
 
     def quit(self):
         """Ask the GUI to quit"""
@@ -94,15 +94,15 @@ class Proxy(object):
     def rise(self):
         """Rise GUI from hidden"""
         self.vessel.show()
-        self._dispatch("rise")
+        return self._dispatch("rise")
 
     def inFocus(self):
         """Set GUI on-top flag"""
-        self._dispatch("inFocus")
+        return self._dispatch("inFocus")
 
     def outFocus(self):
         """Remove GUI on-top flag"""
-        self._dispatch("outFocus")
+        return self._dispatch("outFocus")
 
     def kill(self):
         """Forcefully destroy the process"""
@@ -127,20 +127,6 @@ class Proxy(object):
 
     def validate(self):
         return self._dispatch("validate")
-
-    def _remove_event_filter(self):
-        event_filter = _state.get("eventFilter")
-        if isinstance(event_filter, QtCore.QObject):
-
-            # (NOTE) Should remove from the QApp instance which originally
-            #        installed to.
-            #        This will not work:
-            #        `QApplication.instance().removeEventFilter(event_filter)`
-            #
-            event_filter.parent().removeEventFilter(event_filter)
-            del _state["eventFilter"]
-
-            print("The eventFilter of pyblish-qml has been removed.\n")
 
     def _alive(self):
         """Send pulse to child process
@@ -192,7 +178,6 @@ class Proxy(object):
         except IOError:
             # subprocess closed
             self.vessel.close()
-            self._remove_event_filter()
         else:
             return True
 
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -115,8 +115,12 @@ class Proxy(object):
         self.vessel.setGeometry(x, y, w, h)
         self._dispatch("host_attach")
 
-    def popup(self):
-        self.vessel.activateWindow()  # to top
+    def popup(self, alert):
+        # No hijack keyboard focus
+        QtWidgets.QApplication.setActiveWindow(self.vessel)
+        # Plus alert
+        if alert:
+            QtWidgets.QApplication.alert(self.vessel.parent(), 0)
 
     def publish(self):
         return self._dispatch("publish")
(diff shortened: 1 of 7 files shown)


davidlatwe:
Here's my plugins [Gist](https://gist.github.com/davidlatwe/70e5ed3513f3af4ab1de624b20b63f4a) that used in tests.

mottosso:
Stellar effort @davidlatwe! I've given this a go on Windows 10, Maya 2018 and all went well. There's some flickering in the window when it goes from being a child to its own window, but I can't see a way around that.

I've added some docstrings, but was hoping you could fill in what "ninja" meant, [here](https://github.com/pyblish/pyblish-qml/pull/297/files#diff-b0ba841ec338f7c1732666d737564b5dR231).

mottosso:
Looks like something is off for Maya 2015 and 2016, presumably related to Qt 4 vs. Qt 5.

image

mottosso:
This produces the same result, not sure if it helps.

```python
from pyblish_qml import show
show(foster=False)
```

mottosso:
Also to confirm that the current master, and the master before merging the foster update, works in 2015. So there's something in this PR specifically causing it not to run.

davidlatwe:
Thanks for quick spin @mottosso !

> Looks like something is off for Maya 2015 and 2016, presumably related to Qt 4 vs. Qt 5.

Yes, it's related to Qt version, but the reason it freeze was from the change in commit 7b3277c .
It's about `ninja`, too. `ninja` is foster mode's window detach-attach transition process, which I could not come up with a better name.
I would like to elaborate more, but I need to hit the road now, I'll come back later :)

pyblish_qml/ipc/server.py
```diff
@@ -104,11 +104,12 @@ def show(self, settings=None):
...
-        if not self.ninja:
+        if self.ninja:
+            return self._dispatch("show", args=[settings or {},
+                                                self._winId,
+                                                self.ninja])
+        else:
             self.vessel.show()
-        return self._dispatch("show", args=[settings or {},
-                                            self._winId,
-                                            self.ninja])
```
davidlatwe: This spot, is the reason that make Maya 2015 and 16 freeze at startup. 
`ninja` is disabled if host running Qt4, and if `ninja` is disabled, it need to call `vessel.show` before `_dispatch("show")`, or the window will not show up.

davidlatwe: This spot, is the reason that make Maya 2015 and 16 freeze at startup. 
`ninja` is disabled if host running Qt4, and if `ninja` is disabled, it need to call `vessel.show` before `_dispatch("show")`, or the window will not show up.

davidlatwe:
Okay, just tested on 2015, it should wok now :)

I renamed the param `ninja` to `foster_fixed`, should be clear enough I think. It's to control whether the GUI to avoid main thread busy by parent back to subprocess or not.

If host's `Qt version==4`, GUI will become blank white after window `setParent` back to subprocess from host (tested on Windows). So I made this extra switch `foster_fixed`, to let the fostered window always stay inside the host, although this will produce laggy GUI when host main thread is busy, but it's better then a blank window.

By default, the bool value of `foster_fixed` is set from host's Qt version, one can override it through additional environment variable `PYBLISH_QML_FOSTER_FIXED=True`.
Here's the code that define the value (describe better then my words) :
```python
if not foster:  # If Foster mode is not turned on, `foster_fixed` is False
    return False

value = os.environ.get("PYBLISH_QML_FOSTER_FIXED", "").lower()
return value in ("true", "yes", "1") or QtCore.qVersion()[0] == "4"
```

> There's some flickering in the window when it goes from being a child to its own window, but I can't see a way around that.

If everything works well, this window flickering and the taskbar alert change to main window would be the last two trade off. :P

One other buggy appearance was on CentOS, the window did not resize properly when it parent back to subprocess. But I am going to let it slide for now, will back to fix this one after a while. :)

Does `foster_fixed` works for you ? @mottosso

mottosso:
Works!

Great work @davidlatwe. If you're ready, I'm happy to merge and release. Have you incremented the version too?

mottosso:
Bam! https://github.com/pyblish/pyblish-qml/releases/tag/1.8.2

PR #304, opened 2018-10-01
Enhancement, fix #281, #274 and #119

davidlatwe:
This PR provide 2 fix and 1 extra feature:
* Fix #281: No need to press twice when section items were all manually toggled on/off.
* Fix #274 and #119: Status message are now more specific.
* Able to press stop on collecting, which I think would be a bit improve on UX.

### Before

qml_enhance_before

### After

qml_enhance_after

PR #305, opened 2018-10-02
Fix #304, increase QML initialization waiting time

davidlatwe:
### Problem
In previous PR #304, the new feature that allow user to stop collecting process at startup, will make GUI stuck at "Collecting.." phase if the collecting process ended in *short time* (before QML get initialized).

That feature rely on QML to receive the signal `firstRun` to reset those button, so if the signal `firstRun` emitted before QML complete initialization, GUI stocked.

### Fix
Increase QML initialization waiting time from `500` to `1500`, give it enough time to be ready to receive signal `firstRun`

### DIFF
--- a/pyblish_qml/qml/Overview.qml
+++ b/pyblish_qml/qml/Overview.qml
@@ -208,15 +208,7 @@ Item {
 
         visible: overview.state != "initialising"
 
-        mode: {
-            if (startup === true) {
-                // Ensure collecting process is stoppable on first run
-                setMessage("Collecting..")
-                return 1
-            }
-
-            return overview.state == "publishing" ? 1 : overview.state == "finished" ? 2 : 0
-        }
+        mode:  overview.state == "publishing" ? 1 : overview.state == "finished" ? 2 : 0
 
         width: parent.width
         anchors.bottom: parent.bottom
@@ -238,7 +230,6 @@ Item {
         onFirstRun: {
             app.commentEnabled ? commentBox.up() : null
             commentBox.text = app.comment()
-            footer.startup = false
         }
 
         onStateChanged: {
--- a/pyblish_qml/qml/Overview.qml
+++ b/pyblish_qml/qml/Overview.qml
@@ -209,14 +209,10 @@ Item {
         visible: overview.state != "initialising"
 
         mode: {
-            if (app.state === "collecting" && startup === true) {
+            if (startup === true) {
                 // Ensure collecting process is stoppable on first run
                 setMessage("Collecting..")
                 return 1
-            }
-            if (overview.state == "") {
-                // Make sure the message get displayed on first run
-                setMessage("Ready")
             }
 
             return overview.state == "publishing" ? 1 : overview.state == "finished" ? 2 : 0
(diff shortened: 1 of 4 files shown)


davidlatwe:
With the increased waiting time, the state machine is now able to work right on the first reset,  `Footer.startup` flag and other extra code are no longer needed.

pyblish_qml/qml/Overview.qml
```diff
@@ -208,13 +208,7 @@ Item {
...
         visible: overview.state != "initialising"
 
-        mode: {
-            if (startup == true) {
-                setMessage("Collecting..")
-                return 1
-            }
-            return overview.state == "publishing" ? 1 : overview.state == "finished" ? 2 : 0
-        }
+        mode:  overview.state == "publishing" ? 1 : overview.state == "finished" ? 2 : 0
```
davidlatwe: This was the true bug.

pyblish_qml/qml/Overview.qml
```diff
@@ -208,13 +208,7 @@ Item {
 
         visible: overview.state != "initialising"
 
-        mode: {
-            if (startup == true) {
-                setMessage("Collecting..")
-                return 1
```
davidlatwe: This was the true bug.

BigRoy:
Thanks @davidlatwe - I'll play with this hopefully soon to test it. Just wondering, is the longer 1500 delay still needed if it's fixed already? What's the _current_ worst case?

davidlatwe:
Thanks @BigRoy  :)

> is the longer 1500 delay still needed if it's fixed already?

Not really needed, but would increase the chance that QML event handler get initialized to catch signal `firstRun` if background workload making it slow response.

> What's the *current* worst case?

I think you meant 500 delay ? If so, then the worst case would be the first collecting process will startup with the GUI below, instead of "Collecting.." message and a stop button :

test

No messages and all the buttons were early exposed, not really serious.

BigRoy:
Seems to be working fine on my machine here. Putting this live in production Today, if anything pops up I'll let you know.

davidlatwe:
Thanks @BigRoy !

BigRoy:
It seems to work nicely here.

I did get this comment from one of the artists Today though: 

> I've had pyblish stall on repair actions btw, happened a couple of times today. Not sure why, but it just hangs on 'busy'

That would be the right click menu on plug-ins that trigger an action.
I think it's unrelated to this particular feature (just thought I'd mention it here), but have you seen that happen too maybe?

davidlatwe:
Hmm, No, I haven't.
I do have some repair actions on hand, but didn't hanging in pass few day :)

davidlatwe:
Yep, just checked the change again, should be unrelated, I could bet a case of beer :beer: , haha
Thanks for approval @BigRoy : )
Then I will merge this tomorrow, if there's no other issue pops up.

BigRoy:
> Yep, just checked the change again, should be unrelated, I could bet a case of beer beer , haha

With how often artists are wrong I'd stay away from joining that bet. Nice work again David, thanks. :beers:

davidlatwe:
Merging !

PR #306, opened 2018-10-07
Fix state machine transision bug from PR #304

davidlatwe:
In the state machine, `extracting` state needs to transition to `finished` if the validation failed during the publish, or the stop button won't change to reset button, and cause user operation not able to continue.

qml_fix304_before

This commit fixed that.

qml_fix304_after

---

Somehow I missed this one in the previous tests, after founding this bug in my work, right after the release of `1.8.6`, I spend a day did a series of user operation flow inspection, including each kind of `Action`, and compared the outcome with previous releases, so far this is the only one that I have missed based on previous PRs.

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -253,6 +253,7 @@ class Controller(QtCore.QObject):
         validating.addTransition(self.extracting, extracting)
 
         extracting.addTransition(self.stopping, stopping)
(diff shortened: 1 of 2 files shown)


mottosso:
Fantastic work @davidlatwe, you are welcome to merge!

davidlatwe:
Thanks, merging ! :heart:

PR #312, opened 2018-12-27
Foster Mode Deprecated

davidlatwe:
### Motivation

After introduced experimental feature *Foster Mode* since [`v1.8.0`](https://github.com/pyblish/pyblish-qml/releases/tag/1.8.0), it also spawned a few bugs which related to DCC app's threading, although those bugs seems to be fixed, but the confidence on this feature did not bounce back.

### Implementation

Remove experimental window *fostering* feature from master branch for not being a potential obstacle on further development. The version will bump to `1.9.0`

[Comparing against v1.7.4](https://github.com/pyblish/pyblish-qml/compare/1.7.4...b48efeb)

The feature `Foster Mode` will remain available on my forked branch [`dev-foster`](https://github.com/davidlatwe/pyblish-qml/tree/dev-foster), and will try to update along with the latest master.

Please let me know what you think on this.

### DIFF
--- a/pyblish_qml/__init__.py
+++ b/pyblish_qml/__init__.py
@@ -5,12 +5,12 @@ from .version import (
 )
 
 
-def show(parent=None, targets=None, modal=None, foster=None):
+def show(parent=None, targets=None, modal=None):
     from . import host
 
     if targets is None:
         targets = []
-    return host.show(parent, targets, modal, foster)
+    return host.show(parent, targets, modal)
 
 
 _state = {}
--- a/pyblish_qml/__init__.py
+++ b/pyblish_qml/__init__.py
@@ -5,8 +5,11 @@ from .version import (
 )
 
 
-def show(parent=None, targets=None, modal=None):
+def show(parent=None, targets=None, modal=None, foster=None):
     from . import host
+
+    if foster is not None:
+        print("Foster Mode has been deprecated.")
 
     if targets is None:
         targets = []
--- a/README.md
+++ b/README.md
@@ -105,7 +105,6 @@ See [pyblish-maya](https://github.com/pyblish/pyblish-maya#usage) for an example
 
 **Additional Environment Variables**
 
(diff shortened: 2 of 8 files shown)


pyblish_qml/__init__.py
```diff
@@ -5,12 +5,12 @@
 )
 
 
-def show(parent=None, targets=None, modal=None, foster=None):
+def show(parent=None, targets=None, modal=None):
```
mottosso: Even though `foster=` doesn't do anything, we should keep it around for a few versions, just to ensure it doesn't break people's existing installations.

davidlatwe: I'll put a warning if `foster is not None`
Thanks !

BigRoy:
So what behavior/functionality in particular is this dropping? What did the *foster mode* help to do again? What would the end user see as differing functionality?

davidlatwe:
Thanks for the questions !

> What behavior/functionality in particular is this dropping?

It's dropping the optional ability to embed QML window from subprocess into the host (e.g. Maya), the feature was called `Foster Mode`, since it's *fostering* subprocess window.

> What did the *foster mode* help to do again?

It's making QML window as a child window of the DCC App's main window, instead of a separate window that could hide behind or always on top of. Pure UX stuff which originally aim to make pyblish-qml to behave more like being part of the DCC App (parented window).

See [this gif](https://user-images.githubusercontent.com/3357009/44475434-96676180-a667-11e8-9064-7e404b7425fa.gif)

> Is it just that in Maya for example the UI can never appear behind the Script Editor? (or other separate windows because it now, with the older functionality, always appears in front)

Yes, without `Foster Mode`, it will never get *in between* host main window and any other child window, e.g. Maya main window and Script Editor.
It will either get behind the main window or be on top of all the windows, which driven via an old functionality (introduced in v1.3.0) that implemented with `eventFilter` and window flags.

---

Please let me know if there's still something not clear enough !

BigRoy:
This is perfectly clear - good to have this as a reference in this PR. :)

davidlatwe:
About to merge this !

Issue #281, opened 2018-03-05
Toggling family groups


asztalosdani:
### Description
Clicking on the family gives unexpected results. See attached gif.
![pyblish_qml](https://user-images.githubusercontent.com/9664767/36964268-616f6ef6-2056-11e8-9681-7955e38a3c0b.gif)
### Expected result
1. The group should only uncheck instances if all of them are checked, and check all unchecked ones.
2. Or a proper tri-state checkbox could be implemented.
### How to reproduce
```python
import pyblish.api
class ACollector(pyblish.api.ContextPlugin):
    order = pyblish.api.CollectorOrder - 0.1
    def process(self, context):
        context.create_instance("A instance", family="family")
        context.create_instance("B instance", family="family")
        
pyblish.api.register_plugin(ACollector)
```
1. Uncheck an instance.
2. Click the family -> the unchecked instance is checked.
3. Uncheck both instances.
4. Click the family -> nothing happens
5. Check both instances.
6. Click the family -> nothing happens

mottosso:
Thanks for these detailed reports, @asztalosdani.

I'm also noticing there is a missing icon on the Pyblish window, any idea what that could be?

asztalosdani:
> I'm also noticing there is a missing icon on the Pyblish window, any idea what that could be?
No, not yet. The icon file is there, and pyblish_qml is reading it from the correct place.

PR #314, opened 2019-01-13
Fix 309

### LINKED ISSUE #309: Toggling family groups does not respect mandatory Instances
I discovered that the feature introduced by #281 does not respect whether any of the instances in that group have `.data['optional']` set to `False`, ie "mandatory" or "non-optional" instances, which I had introduced support for as of version 1.7.4.

Toggling should not fail, but it should skip the non-optional Instances from being toggled.

davidlatwe:
This PR will fix #309 , and UX bug related to #281 but on plugin section.

### Before
qml_309

* The instance `Batman` and `David` should not be toggled, but the restriction got bypassed via toggling the section.
* Plugin checkbox did not give immediate respond to section toggling.

### After
qml_309fix

* All fixed !

### Implementation

1. Collect all *optional* items' state.
2. Override section state if all optional items are in the same state. This guaranteed the optional item be able to respond to every toggle.
3. Set check state.

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -507,6 +507,7 @@ class Controller(QtCore.QObject):
             if item.itemType == "instance" and sectionLabel == item.category:
                 if item.optional:
                     states.add(item.isToggled)
+
             if item.itemType == "plugin" and sectionLabel == item.verb:
                 if item.optional:
                     states.add(item.isToggled)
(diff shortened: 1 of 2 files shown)


mottosso:
Nice work! Could you increment the version too? Then I'll make a new release for it.

davidlatwe:
Done !

BigRoy:
Good stuff again David! :+1:

tokejepsen:
Awesome work @davidlatwe

davidlatwe:
Thanks :)
Merging this, and pinging @mottosso for release :P

mottosso:
Yes, that is all fine! :)

One question though, that I noticed elsewhere as well, how come the instances are green, and yet checkable? :O I thought they were green only after having been processed, at which point they would not be checkable anymore?

mottosso:
Oh and I remembered you are able to make releases too, you are most welcome to do that here as well!

davidlatwe:
> how come the instances are green, and yet checkable?

The reason it's green is that there's a plugin printing log message for every instances in collecting phase.
```python
class ListHeros(pyblish.api.InstancePlugin):
    order = pyblish.api.CollectorOrder + 0.1

    def process(self, instance):
        self.log.info(instance)
```
So yeah, they did been processed :) 

> at which point they would not be checkable anymore?

I intentionally set `data["optional"] = False` to some instances at their creation, for the example ;D

> Oh and I remembered you are able to make releases too

Haha, thanks :D

BigRoy:
> One question though, that I noticed elsewhere as well, how come the instances are green, and yet checkable? :O I thought they were green only after having been processed, at which point they would not be checkable anymore?

*AND*

> The reason it's green is that there's a plugin printing log message for every instances in collecting phase.

Is this related to https://github.com/pyblish/pyblish-qml/issues/283?

Issue #315, opened 2019-01-14
Continually reflecting the changes of instance's data to GUI


davidlatwe:
## Motivation

Here's my scenario, I want to publish a turntable rendering, but I also need to be sure that the asset has been publish and not been changed.

So I design the turntable publishing workflow like this:
1. Artist publish his/her asset
2. Pyblish plugin lock the Maya scene after integration
3. Artist publish turntable

**And to be more secure, I want to force deactivate turntable rendering instance if scene is not locked, and deactivate other asset instance after scene is locked.**

But I found that, QML does not support instance data update and reflect the change to the GUI, after the instance been created.

## Problem

Here's the GIF

![qml_notupdate](https://user-images.githubusercontent.com/3357009/51100106-751e3200-180f-11e9-9c85-4212b8b06181.gif)

In the GIF above:
* The plugin **`SetupTurntable`** will add family `imagesequence.turntable` into the *Render* instance, and deactivate the instance if scene is not locked.
* The plugin **`DeactivateOnSceneLock`** will deactivate all instances if the scene is locked, unless the instance has `"_protect_on_lock"` key in `data`.
* The plugin **`ShowActive`** prints out the instance that it processed.

So far, the data I have changed during the collection phase were:
* `optional`, but the instance still toggleable after I set this to `False`
* `publish`, but the instance did not toggle off when I set this to `False`
* `families`, but the perspective view did not show the newly added family `imagesequence.turntable`
<br></br>
<details>
 <summary>Here's the code</summary>

```python

import os
import pyblish.api


instances = {
    "Render": {
        "family": "imagesequence",
    },
    "Geometry": {
        "family": "model",
    },
}


def scene_locked():
    """Scene lock simulator"""
    return bool(os.environ.get("SCENE_LOCK"))


def deactivate_instance(cls, instance):
    instance.data["optional"] = False
    instance.data["publish"] = False
    cls.log.info("Instance {!r} blocked !".format(instance.name))


class CreateInstances(pyblish.api.ContextPlugin):
    order = pyblish.api.CollectorOrder - 0.1

    def process(self, context):
        for name, data in instances.items():
            instance = context.create_instance(name)
            instance.data.update(data)


class SetupTurntable(pyblish.api.InstancePlugin):
    """Setup turntable rendering instance

    Must wait until `model` has been published

    """
    order = pyblish.api.CollectorOrder
    families = ["imagesequence"]

    def process(self, instance):
        # Setup protection flag
[...]

PR #316, opened 2019-01-15
Implementing #315

davidlatwe:
The implementation of #315, which aim to keep the GUI to display the latest status data change of instances, like `optional`, `publish`, `families`.

Currently, once the plugin which responsible to create the instance, has been processed, the change of the instance data that made by other plugins will not reflected to GUI. For example, the newly added family will not be shown in Perspective View, and setting `publish` to False will not toggle instance off.

Not good to merge yet.

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -991,6 +991,8 @@ class Controller(QtCore.QObject):
 
             self.data["models"]["item"].update_with_result(result)
             self.data["models"]["result"].update_with_result(result)
+            # Update proxy instance data which currently being iterated in
+            # the primary iterator
             update_instance_with_result(result)
 
             # Once the main thread has finished updating
@@ -1006,8 +1008,7 @@ class Controller(QtCore.QObject):
             proxy = next((i for i in context if i.id == id), None)
             if proxy is None:
                 return
-            # Update proxy instance data which currently being iterated in
-            # the primary iterator
+
             proxy.data["publish"] = data.get("publish", True)
             proxy.data["family"] = data["family"]
             proxy.data["families"] = data.get("families", [])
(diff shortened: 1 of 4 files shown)


davidlatwe:
I have completed the implementation of GUI's instance status update and instance removal, this will make pyblish-qml's primary iterator behaves like pyblish-base's.

Which means plugins could change instance's `family`, `publish` status data or removing instance from context, and affect the publish flow in run-time with GUI.

### Implementations

There are two kinds of instances in pyblish-qml's GUI process, instance model item used by GUI for displaying, and proxy instances that are being iterated in the primary iterator. And they all get updated by the cached instances which parsed from host at each iteration, but with slightly difference.

For presenting correct info on GUI, instance model item needs to update these attributes:
  - `isToggled`
  - `optional`
  - `category`
  - `familiesConcatenated`

For iterating instances correctly with GUI's iterator, the instance proxy needs to update these data entries:
  - `data["publish"]`
  - `data["family"]`
  - `data["families"]`

And in the primary iterator, it's will skip the instance that has been marked as removed or the publish flag has been switched off.

---

I think it's good to be reviewed now, please let me know what you think about this enhancement.

Thanks !

asztalosdani:
Hi,

we have a few plugins that update the `families` and `publish` flags, so it would be good to see this merged. :)

mottosso:
Happy to merge this, @davidlatwe when you're happy I'm happy!

davidlatwe:
Thanks!
Have been running this in production for a while here, I’ll take a look at this again and merge this later if all okay !

davidlatwe:
I have pushed two more commits that was only in my local production branch, one is an improvement on instance removal (d50e7e9), and the other (3082f8a) was for debugging (Sorry for mixing this one here).

Please let me know if there's any problem :)

davidlatwe:
Merging this :)

PR #320, opened 2019-02-14
No Qt

mottosso:
This PR removes the Qt dependency from a host, such that only the external Python process requires it.

Before, Qt was imported regardless of host, to try and show a splash screen, install an event filter and listen on the application quit signal. Now that only happens if Qt is available.

The Qt import mechanism itself was moved into a closed scope, such that it leaves mock-functions available for hosts that lack access to Qt. The result is no loss in functionality, and no need for PyQt5 to exist from a given host. Now all you need is love. And Python.

### DIFF
--- a/pyblish_qml/host.py
+++ b/pyblish_qml/host.py
@@ -1,15 +1,17 @@
 import os
 import sys
 import inspect
+import logging
 import traceback
 
 import pyblish.api
 
 from . import ipc, settings, _state
-from .vendor.Qt import QtWidgets, QtCore, QtGui
 
 MODULE_DIR = os.path.dirname(__file__)
 SPLASH_PATH = os.path.join(MODULE_DIR, "splash.png")
+
+log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


mottosso:
Oh, and it also removes the need for PyQt5 to exist on PYTHONPATH, which can complicate this scenario:

1. Pyblish QML uses Python 3, from a Python 2 host, e.g. Maya 2018
2. PyQt5 is installed to the target Python 3's site-packages

In the above case, one would need to append Python 3's site-packages to the PYTHONPATH of the given host, which would give Python 2 access to its modules. The "proper" was of handling this is to not install PyQt5 in site-packages, but elsewhere, e.g. `/special/pyqt5/path` such that only that path could be added to PYTHONPATH.

Now you don't need to do that either.

tokejepsen:
Very cool stuff! Can't review it properly atm.

Would make it much easier to package and have consistent behaviour through all apps.

mottosso:
Taking this for a spin in production.

PR #322, opened 2019-02-15
Windows Bug Fixes

mottosso:
Without `PYBLISH_QML_PYQT5`, the wrong path got automatically registered causing Qt binaries to be exposed to the Pyblish QML process (e.g. PySide2 binaries for a PyQt5 application).

Also when the above error occurred, the event filter was supposed to get uninstalled, but was not. That's fixed here too.

### DIFF
--- a/pyblish_qml/host.py
+++ b/pyblish_qml/host.py
@@ -268,7 +268,7 @@ class QtHost(Host):
         }
 
         class EventFilter(QtCore.QObject):
-            def eventFilter(self, widget, event):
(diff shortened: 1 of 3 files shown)


davidlatwe:
Haha, just about to report that the `eventFilter` was not working, but I can confirm that has been fixed !
Thanks :)

mottosso:
Thanks for confirming. :)

mottosso:
Taking this for another spin.

PR #323, opened 2019-02-16
Always try to install on show

davidlatwe:
### Bug fix
This fixes the bug that QML only install eventFilter once on first run :)
(Missed this one because I only tried running it once in Maya while testing previous PR)


mottosso:
Thanks David, missed this one!

PR #326, opened 2019-04-11
Some QoL features for integration

nasefbasdf:
These are some features to streamline the usage of pyblish-qml in our pipeline. They are implemented backwards compatible,  no previous usage/behaviour is expected to break.

Basically we have our own UI that deals with all the instance/plugin config. This means when we call the pyblish-qml app, everything is already set up. Our changes are related to automation:

1. One of the added features is a new argument to `show(... , auto_publish_at_first_run=False)`. This saves the user the wait for the reset and the click on the publish button.

1. Another added feature is also a new argument to `show(... , comment=None)`. In our system the user needs to input the comment earlier than the pyblish process can be started, so we wish to be able to pass it along to pyblish-qml.

1. Lastly, two API entry points were added: `host.hide` and `host.quit`. These were available on the `Proxy` object but they were not readily accessible like `host.publish` or `host.validate`. We use these to automatically exit the UI if all processes succeed.

### DIFF
--- a/pyblish_qml/host.py
+++ b/pyblish_qml/host.py
@@ -75,7 +75,7 @@ def uninstall():
     sys.stdout.write("Pyblish QML shutdown successful.\n")
 
 
-def show(parent=None, targets=[], modal=None):
+def show(parent=None, targets=[], modal=None, comment=None, auto_publish_at_first_run=False):
     """Attempt to show GUI
 
     Requires install() to have been run first, and
@@ -95,13 +95,17 @@ def show(parent=None, targets=[], modal=None):
     # Automatically install if not already installed.
     install(modal)
 
+    show_settings = settings.to_dict()
+    show_settings['comment'] = comment
+    show_settings['autoPublishAtFirstRun'] = auto_publish_at_first_run
+
     # Show existing GUI
     if _state.get("currentServer"):
         server = _state["currentServer"]
         proxy = ipc.server.Proxy(server)
 
         try:
-            proxy.show(settings.to_dict())
+            proxy.show(show_settings)
             return server
 
         except IOError:
@@ -120,7 +124,7 @@ def show(parent=None, targets=[], modal=None):
         return host.desplash()
 
     proxy = ipc.server.Proxy(server)
-    proxy.show(settings.to_dict())
+    proxy.show(show_settings)
 
     # Store reference to server for future calls
     _state["currentServer"] = server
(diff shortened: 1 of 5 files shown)


mottosso:
Thanks for this!

> 1 ... show(... , auto_publish_at_first_run=False)

This is a great feature, but what do you think about the verbose argument name? How about `auto_publish` and to complete it a `auto_validate` as well? It's pretty self-explanatory (I think?) that it'll only happen on the first run, as `show()` is only called once... Or is it?

> 2 ... In our system the user needs to input the comment earlier than the pyblish process can be started, so we wish to be able to pass it along to pyblish-qml.

It's not terribly obvious, but are you aware of the `context.data["comment"]` field? That's there the "initial state" of the comment text field is coming from. You can pre-populate this field using a Collector.

nasefbasdf:
Alright, I'll have a look at and try to include auto_validate as well. However, it will only be really useful if the user can continue after validation. Right now, the only option after validation is to reset. I won't be able to look into changing that right now.

These "auto" flags will only initiate procedures *once* after show is called. I effectively reused the logic around `self.firstRun.emit()`.

I've just realized how the comment must be passed in. A new plugin is needed to pull the comment from our systems and put it on the pyblish context. I'll revert that part of the commit.

pyblish_qml/host.py
```diff
@@ -158,6 +162,32 @@ def validate():
             # The running instance has already been closed.
             _state.pop("currentServer")
 
+            
+def hide():
+    # get existing GUI
+    if _state.get("currentServer"):
```
mottosso: Should we throw an error here, if there isn't a `currentServer`, to let the caller know it didn't work? At the moment, it looks like it would silently fail.

davidlatwe: I think we could let this slide for now, because these two methods `host.validate` and `host.publish` in above lines work in the same fashion ( no error if no `currentServer`).
Should be in another PR to resolve them all together.

nasefbasdf: I've essentially duplicated the code from `host.publish`. I've not thought deeply into how it works or how it doesn't work.
Since validate, publish, hide, quit (and possibly others later) share most of the code here, it would be better to create a decorator.
```
from functools import wraps
...

def proxy_call(func):
    @wraps(func)
    def proxy_call_wrapper(*args, **kwargs):
        # get existing GUI
        if _state.get("currentServer"):
            server = _state["currentServer"]
            proxy = Proxy(server)
            try:
                return func(proxy, *args, **kwargs)
            except IOError:
                # The running instance has already been closed.
                _state.pop("currentServer")
    return proxy_call_wrapper
...

@proxy_call
def hide(proxy):
    proxy.hide()
```
Or something like that. Error handling can be added once it is decided what kind should be there of course.

mottosso: Yes, I think that makes sense. Happy with that.

mottosso:
Our of personal curiosity, are you free to mention what studio are you integrating this for?

nasefbasdf:
I work for a software developer company called Limes based in Budapest, Hungary. Our asset and project management solution also called Limes, and our primary client is Puppetworks Animation Studio.

mottosso:
Happy with this. @nasefbasdf when you're happy, please increment the version and I'll make a new release.

nasefbasdf:
I've included the `proxy_call` decorator discussed above and incremented the version. This is done from my point of view.

mottosso:
Thanks @nasefbasdf!

PR #329, opened 2019-04-17
Re-apply #325 and fix #327

### LINKED ISSUE #327: Empty lines printed
Hi,

the fix in #325 introduced a funky side-effect.
A ton of empty lines get printed to stdout (the output window of Maya). It has to be something with the two `\n` characters not parsed so printed to sdtout instead.
You can check it by typing something into the comment field of the GUI. For every key pressed a newline (or maybe some?) is printed.

@nasefbasdf, @mottosso can you take a look at it? :)

nasefbasdf:
This contains a fix for #327 while re-applying #325 .

The async consumption of a sync queue discovered during the investigation of #327 is a separate problem:

- This fix ensures different threads on the client side that print will not mess up the communication.
- While the async consumption of messages is a design issue where two legitimate communication procedures may encounter a race condition.

I propose that a new issue should be opened for the async problem.

### DIFF
--- a/pyblish_qml/ipc/client.py
+++ b/pyblish_qml/ipc/client.py
@@ -168,7 +168,10 @@ class Proxy(object):
         assert self.channels["response"].empty(), (
             "There were pending messages in the response channel")
 
-        sys.stdout.write(data + "\n")
+        # To ensure successful IPC message parsing, the message and the
+        # surrounding delimiters must be passed to the stream object at once.
+        # See https://github.com/pyblish/pyblish-qml/pull/325 for more info.
+        sys.stdout.write("\n" + data + "\n")
         sys.stdout.flush()
 
         try:
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -260,8 +260,6 @@ class Server(object):
                         sys.stdout.write(line)
 
                 else:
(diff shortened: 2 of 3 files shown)


mottosso:
Thanks for this (and your patience) @nasefbasdf.

Are you able to test this on your end, preferably in production, prior to merging?

nasefbasdf:
I've tested it.

The deadlock issue was easily reproducible with different timings of the threads. It came up rarely in production, but accidentally I've found a break-point combination where it would do it all the time during debugging.

I've checked and the empty lines are not printed with this change.

mottosso:
2 weeks later, if everything is still OK then it's merging time! :D

nasefbasdf:
I've failed to properly cover one case where the buffered newline should be printed. (when the line is parsed as a JSON, but it's not actually a pyblish message) That's fixed by the last commit.

Otherwise I've been using the modified version without problems.

mottosso:
Thanks @nasefbasdf!

BigRoy:
I *just* updated to the latest version of QML and I think due to this PR the Maya script editor gets spammed with tons of empty lines.

- [Reference](https://gitter.im/pyblish/pyblish?at=5cf11c1e9b50f639355b5ad3)

Any idea why it might not work well in my scenario?

_Windows 10 + Maya 2018_

PR #330, opened 2019-05-16
WindowsError raised in Houdini after #320

davidlatwe:
### Problem

After PR #320, `WindowsError: [Error 6]` raised in Houdini while `PyQt5` lookup at this line [here](https://github.com/pyblish/pyblish-qml/blob/e93b144f7c0bf7ea5606e8449399d77ef4943753/pyblish_qml/ipc/server.py#L392).
I changed from using `subprocess.check_output` to `Popen` still getting same error.
It seems something to do with `subprocess.PIPE`.

### Solution

Not sure if this is the right fix, but for what I can do now is adding `WindowsError` exception.

---

### Update
The main issue was we need both `stdin` and `stdout` set to `subprocess.PIPE` in order to work inside Houdini ( without Administrator privilege ).
So here's the main changes for working in Houdini:
* Add `stdin=subprocess.PIPE`
* Run `check_output` with clean environment

### DIFF
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -399,7 +399,6 @@ def find_pyqt5(python):
                 ],
                 universal_newlines=True,  # Normally, the output is bytes.
                 stdin=subprocess.PIPE,
-                env={"SYSTEMROOT": os.getenv("SYSTEMROOT"), "PATH": ""},
                 creationflags=CREATE_NO_WINDOW)
 
             pyqt5 = os.path.dirname(os.path.dirname(path))
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -397,7 +397,7 @@ def find_pyqt5(python):
                 "sys.stdout.write(PyQt5.__file__)"
 
                 # Normally, the output is bytes.
-            ], universal_newlines=True)
+            ], universal_newlines=True, stdin=subprocess.PIPE)
 
             pyqt5 = os.path.dirname(os.path.dirname(path))
 
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -391,13 +391,16 @@ def find_pyqt5(python):
     # such as Python 2.
     if not pyqt5:
         try:
-            path = subprocess.check_output([
-                python, "-c",
-                "import PyQt5, sys;"
-                "sys.stdout.write(PyQt5.__file__)"
-
-                # Normally, the output is bytes.
-            ], universal_newlines=True, stdin=subprocess.PIPE)
+            path = subprocess.check_output(
+                [
+                    python, "-c",
+                    "import PyQt5, sys;"
+                    "sys.stdout.write(PyQt5.__file__)"
+                ],
+                universal_newlines=True,  # Normally, the output is bytes.
+                stdin=subprocess.PIPE,
+                env={"SYSTEMROOT": os.getenv("SYSTEMROOT"), "PATH": ""},
+                creationflags=CREATE_NO_WINDOW)
 
             pyqt5 = os.path.dirname(os.path.dirname(path))
 
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -401,7 +401,7 @@ def find_pyqt5(python):
 
             pyqt5 = os.path.dirname(os.path.dirname(path))
 
-        except (subprocess.CalledProcessError, WindowsError):
+        except subprocess.CalledProcessError:
             pass
 
     return pyqt5
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -129,9 +129,9 @@ class Server(object):
         is_embedded = os.path.split(sys.executable)[-1].lower() != "python.exe"
 
         python = python or find_python()
-        pyqt5 = pyqt5 or find_pyqt5(python)
-
         print("Using Python @ '%s'" % python)
+
+        pyqt5 = pyqt5 or find_pyqt5(python)
         print("Using PyQt5 @ '%s'" % pyqt5)
 
         # Maintain the absolute minimum of environment variables,
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -233,10 +233,11 @@ class Server(object):
         def _listen():
             """This runs in a thread"""
             HEADER = "pyblish-qml:popen.request"
-            
-            # To ensure successful IPC message parsing, the message got a delimiter newline in front
-            # of it. To differentiate between real newlines and message preambles we need to buffer
-            # them until the next part arrives.
+
+            # To ensure successful IPC message parsing, the message got a
+            # delimiter newline in front of it. To differentiate between
+            # real newlines and message preambles we need to buffer them
+            # until the next part arrives.
             last_msg_newline = False
 
             for line in iter(self.popen.stdout.readline, b""):
@@ -253,7 +254,8 @@ class Server(object):
                         last_msg_newline = False
 
                     if line == "\n":
-                        # buffer and print newlines only if they are not preambles of messages
+                        # buffer and print newlines only if they are not
+                        # preambles of messages
                         last_msg_newline = True
                     else:
                         # This must be a regular message.
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -399,7 +399,7 @@ def find_pyqt5(python):
 
             pyqt5 = os.path.dirname(os.path.dirname(path))
 
-        except subprocess.CalledProcessError:
+        except (subprocess.CalledProcessError, WindowsError):
             pass
 
     return pyqt5
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -385,27 +385,6 @@ def find_pyqt5(python):
         os.getenv("PYBLISH_QML_PYQT5")
     )
 
-    # If not registered, ask Python for it explicitly
-    # This avoids having to expose PyQt5 on PYTHONPATH
-    # where it may otherwise get picked up by bystanders
-    # such as Python 2.
-    if not pyqt5:
-        try:
-            path = subprocess.check_output(
-                [
-                    python, "-c",
-                    "import PyQt5, sys;"
-                    "sys.stdout.write(PyQt5.__file__)"
-                ],
-                universal_newlines=True,  # Normally, the output is bytes.
-                stdin=subprocess.PIPE,
-                creationflags=CREATE_NO_WINDOW)
-
-            pyqt5 = os.path.dirname(os.path.dirname(path))
-
-        except subprocess.CalledProcessError:
-            pass
-
     return pyqt5
 
 
--- a/pyblish_qml/version.py
+++ b/pyblish_qml/version.py
@@ -1,7 +1,7 @@
 
 VERSION_MAJOR = 1
 VERSION_MINOR = 9
-VERSION_PATCH = 10
+VERSION_PATCH = 11
 
 version_info = (VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH)
 version = '%i.%i.%i' % version_info


BigRoy:
Quickly looking over [this topic](https://stackoverflow.com/questions/50152689/subprocess-check-output-gives-windows-error) it sounds like it should happen when the command being run through `subprocess` can't be found by Windows. _Which I also believed should only be the case if you say `shell=True` but I'm not sure._

Anyway, can you check whether the Python it finds is actually a valid path for subprocess? e.g. can you bring the issue down to a reproducable `subprocess` call with the same values?

tokejepsen:
I had this issue with Houdini before; https://github.com/bumpybox/pyblish-bumpybox/blob/4f8db925866eddf63788e801351aabdfedb5478f/pyblish_bumpybox/plugins/extract_review.py#L98

davidlatwe:
@BigRoy 
> Anyway, can you check whether the Python it finds is actually a valid path for subprocess?

Yeah, did check this one, the Python it found is valid.
Here's the full log :)
```
Setting up Pyblish QML in Houdini
uninstalling..
Installing..
aboutToQuit..
event filter..
Using Python @ 'C:\miniconda3\envs\avalon\python.exe'
  File "C:\Users\david\Dropbox\github\AVALON\pyblish-qml\pyblish_qml\host.py", line 121, in show
    server = ipc.server.Server(service, targets=targets, modal=modal)
  File "C:\Users\david\Dropbox\github\AVALON\pyblish-qml\pyblish_qml\ipc\server.py", line 134, in __init__
    pyqt5 = pyqt5 or find_pyqt5(python)
  File "C:\Users\david\Dropbox\github\AVALON\pyblish-qml\pyblish_qml\ipc\server.py", line 401, in find_pyqt5
    ], universal_newlines=True)
  File "C:\PROGRA~1\SIDEEF~1\HOUDIN~1.173\python27\lib\subprocess.py", line 216, in check_output
    process = Popen(stdout=PIPE, *popenargs, **kwargs)
  File "C:\PROGRA~1\SIDEEF~1\HOUDIN~1.173\python27\lib\subprocess.py", line 386, in __init__
    errread, errwrite), to_close = self._get_handles(stdin, stdout, stderr)
  File "C:\PROGRA~1\SIDEEF~1\HOUDIN~1.173\python27\lib\subprocess.py", line 519, in _get_handles
    p2cread = self._make_inheritable(p2cread)
  File "C:\PROGRA~1\SIDEEF~1\HOUDIN~1.173\python27\lib\subprocess.py", line 570, in _make_inheritable
    _subprocess.DUPLICATE_SAME_ACCESS)
WindowsError: [Error 6] 
```

@tokejepsen 

I changed to use `Popen` but still getting the same error when I set `stdout=subprocess.PIPE` :/

BigRoy:
Do these error too?

```python
import subprocess
python = r'C:\miniconda3\envs\avalon\python.exe'
# error?
print("1")
subprocess.check_output(python)
print("2")
subprocess.check_output(python, "-c", "import PyQt5, sys; sys.stdout.write(PyQt5.__file__)")
```

And how about enforcing forward slashes on the python environment variable?

---

Also, I've seen some weird things between running Houdini in Administrator mode or not - like with licensing and some other tiny things... so maybe, if you dare, does it also happen when you run Houdini as administrator? Likely the same, just wondering

- Sorry, just curious to see why this is happening only in Houdini.

davidlatwe:
> Do these error too?

Yes :(

But running Houdini in Administrator mode did not raise `WindowsError`  0.0
And returns the pyqt5 path !
 
Here's my changed test code:
```python
import subprocess, os
python = r'C:\miniconda3\envs\avalon\python.exe'
cmd = 'import PyQt5, sys; sys.stdout.write(PyQt5.__file__)'
path = subprocess.check_output([python, "-c", cmd],
                               universal_newlines=True,
                               # Clear out the current env or it will get Houdini's Python lib
                               env={"SYSTEMROOT": os.getenv("SYSTEMROOT"), "PATH": ""})
print(path)
```

davidlatwe:
Ok..
Turns out, without Administrator privilege, the `Popen` still workable, but `check_output` does not.

Here's the workable `Popen` code that runs *without* Administrator privilege :
```python
import subprocess, os

python = r'C:\miniconda3\envs\avalon\python.exe'
cmd = 'import PyQt5, sys; sys.stdout.write(PyQt5.__file__)'
env = {"SYSTEMROOT": os.getenv("SYSTEMROOT"), "PATH": ""}

popen = subprocess.Popen(
    [python, "-c", cmd],
    env=env,
    universal_newlines=True,
    stdin=subprocess.PIPE,
)
print(popen.communicate()[0])
```

And here's the code that *NOT* able to work *without* Administrator privilege,
but *works* if the privilege was given :
```python
import subprocess, os

python = r'C:\miniconda3\envs\avalon\python.exe'
cmd = 'import PyQt5, sys; sys.stdout.write(PyQt5.__file__)'
env = {"SYSTEMROOT": os.getenv("SYSTEMROOT"), "PATH": ""}

popen = subprocess.Popen(
    [python, "-c", cmd],
    env=env,
    universal_newlines=True,
    stdout=subprocess.PIPE,  # <--- It's `stdout`, not `stdin`
)
print(popen.communicate()[0])
```

davidlatwe:
Okay, so the main problem was we need both `stdin` and `stdout` set to `subprocess.PIPE` in order to work inside Houdini ( without Administrator privilege ).
And the `check_output` already used the `stdout` internally, all we need was set the `stdin` as well.
Here's the workable `check_output` code:
```python
import subprocess, os
python = r'C:\miniconda3\envs\avalon\python.exe'
cmd = 'import PyQt5, sys; sys.stdout.write(PyQt5.__file__)'
path = subprocess.check_output([python, "-c", cmd],
                               universal_newlines=True,
                               # Clear out the current env or it will get Houdini's Py2 lib
                               env={"SYSTEMROOT": os.getenv("SYSTEMROOT"), "PATH": ""},
                               stdin=subprocess.PIPE)  # Add stdin
print(path)
```

tokejepsen:
That is some excellent detective work!

davidlatwe:
I think it's good to merge now, also updated the first comment. :smiley:

BigRoy:
@davidlatwe was this `{"SYSTEMROOT": os.getenv("SYSTEMROOT"), "PATH": ""}` necessary to get it to work? It seems like a bit of a hacky workaround for something. Maybe we should add a little comment above that too to describe what it's for. This is something I think we will tend to lose sight on or forget at some point since it's so obscure.

davidlatwe:
That `{"SYSTEMROOT": os.getenv("SYSTEMROOT"), "PATH": ""}` thing was mainly trying to avoid the [`SyntaxError` crash (*NOT CodecRegistryError*)](https://stackoverflow.com/questions/29296374/codecregistryerror-while-running-blender-from-python-script), when the subprocess Python is version 3.

> However, even with this change - wouldn't it still hit that same wall a bit later on with: https://github.com/pyblish/pyblish-qml/blob/543afae5afc9542c013985d425956aee4c972508/pyblish_qml/ipc/server.py#L139
So even if it would check pyqt5 with a changed environment then... in the end it would still run with the local environment - yes?

Yes, I think you are right.
The propose of this subprocess call was meant to ask that Python to find PyQt5 by itself.
And this environment wipe out just defeated it !

I think the right thing to do here for Houdini was simply set the environ var `PYBLISH_QML_PYQT5`.

mottosso:
> And this environment wipe out just defeated it !

Not entirely; it'll still find it's own `site-packages/` directory where PyQt5 may be installed.

davidlatwe:
> I think the right thing to do here for Houdini was simply set the environ var `PYBLISH_QML_PYQT5`.

If that's the case, then I think the purpose of this PR has changed to only patch the subprocess I/O stream issue in Houdini.

At least `SyntaxError` is way better then `WindowsError: [Error 6]`.  :laughing:

davidlatwe:
> it'll still find it's own site-packages/ directory where PyQt5 may be installed.

Oh, yes. But I was thinking that what if PyQt5 was in `sys.path` or any other magical place :/

BigRoy:
Honestly - the autofinding of PyQt5 - does it really do much I wonder? :) Technically it runs the subprocess that it would do later anyway and ask "hey do you have it?" - why not let the subprocess just fail and let that one report it? I noticed when running over a slow network this query alone loading up a Python over the network can take over a second (worst case) and usually takes about 0.6 seconds.

Or in which case does this "search" actually help specifically?

davidlatwe:
> why not let the subprocess just fail and let that one report it?

Hmmm, looks like it provide a way that you won't need to put PyQt5 in `PYTHONPATH` in your working environment as long as the subprocess Python able to find it.
And once it founds it, QML will append it with other env vars to assemble a clean environment.

But I think I'll remove that `env={"SYSTEMROOT": os.getenv("SYSTEMROOT"), "PATH": ""}` stuff for now, since it's looks like a bit off the main topic of this PR (mainly to resolve the `WindowsError`) ?

BigRoy:
> Hmmm, looks like it provide a way that you won't need to put PyQt5 in PYTHONPATH in your working environment as long as the subprocess Python able to find it.

Exactly! Wouldn't that mean that exactly the Python process you will be running in anyway found it? So even if you didn't explicitly find it then it would still have it?

Or @mottosso what am I misunderstanding of the behavior that was implemented? What does it solve and how?

davidlatwe:
> So even if you didn't explicitly find it then it would still have it?

🤔  Oh, it would not be able to find it, if for some reason you already messed up the current environment !

mottosso:
I can't remember, but if you follow the breadcrumbs:

- [Line](https://github.com/pyblish/pyblish-qml/blob/543afae5afc9542c013985d425956aee4c972508/pyblish_qml/ipc/server.py#L388)
- [Commit](https://github.com/pyblish/pyblish-qml/commit/31870fbd920927562d7459fe9f69f1cfe0b45bf9)
- [PR](https://github.com/pyblish/pyblish-qml/pull/320)

It looks like it's related to Blender.

- https://github.com/pyblish/pyblish-qml/pull/320#issuecomment-463694494

davidlatwe:
Looks like the actual commit that related to Blender directly was this one: 9154ed7
That commit solved the problem that Blender does not have access to any Qt binding.

And currently we are looking at this: 31870fb
Which I agreed with @BigRoy that there's no need to do this extra `subprocess query`, looks like the only benefit from there was able to print out the PyQt5 path that will be used, at [here](https://github.com/pyblish/pyblish-qml/blob/fe67a98273eaf2cc014717c1e012e4febe71f378/pyblish_qml/ipc/server.py#L135).

I think we can safely revert it. What do you think ?

mottosso:
I can't say for sure, not without testing with Blender which I don't have setup at the moment.

If it's just to save on performance, I would leave it in. You can optimise and avoid it being called by setting the environment variables explicitly.

davidlatwe:
> If it's just to save on performance, I would leave it in. You can optimise and avoid it being called by setting the environment variables explicitly.

Yes, currently I just put a random string like `hello` to `PYBLISH_QML_PYQT5` :P

BigRoy:
> Yes, currently I just put a random string like hello to PYBLISH_QML_PYQT5 :P
> Since I installed the PyQt5 in client side Python's site-packages

Haha - the hack. I'm for reverting to without this too if it really does "nothing" helpful aside of debug printing where it's at. If you want that debug message I'd say it's best to have Pyblish QML itself (as it launches in the subprocesses) to output what it's importing instead of a separate subprocess.

> I can't say for sure, not without testing with Blender which I don't have setup at the moment.

Any way we can test this? 

I was able to run QML even prior the "blender integration" PR because I installed `PyQt5` with it. Does testing only mean we should be able to run QML even without installing PyQt5 (or any Qt in Blender) - if that's the case then this would be good to go?

mottosso:
I think if this works, we should be good to go.

```bash
$ set PYBLISH_QML_PYTHON_EXECUTABLE=c:\python36  # With PyQt5 installed in site-packages
$ blender
```

You should be seeing a File -> Publish menu item, with QML ready to go.

BigRoy:
### Blender 2.8 Beta

Tried it in Blender 2.8 but no luck getting the file menu. Also not on `import pyblish_qml` or `import pyblish_qml.host`. Then I ran `pyblish_qml.show()` giving me this error:
```python
>>> pyblish_qml.show()
Error: Traceback (most recent call last):
  File "D:\git_pipeline\avalon\development\git\pyblish-qml\pyblish_qml\host.py", line 659, in execute
    self._timer = wm.event_timer_add(0.01, context.window)
TypeError: WindowManager.event_timer_add(): required parameter "window" to be a keyword argument!

location: D:\blender\blender-2.80.0-git.85322737baec-windows64\2.80\scripts\modules\bpy\ops.py:200


  File "<blender_console>", line 1, in <module>
  File "D:\git_pipeline\avalon\development\git\pyblish-qml\pyblish_qml\__init__.py", line 16, in show
    return host.show(parent, targets, modal, auto_publish, auto_validate)
  File "D:\git_pipeline\avalon\development\git\pyblish-qml\pyblish_qml\host.py", line 97, in show
    install(modal)
  File "D:\git_pipeline\avalon\development\git\pyblish-qml\pyblish_qml\host.py", line 68, in install
    install_host(use_threaded_wrapper)
  File "D:\git_pipeline\avalon\development\git\pyblish-qml\pyblish_qml\host.py", line 242, in install_host
    install(use_threaded_wrapper)
  File "D:\git_pipeline\avalon\development\git\pyblish-qml\pyblish_qml\host.py", line 672, in _install_blender
    bpy.ops.wm.pyblish_qml_timer()
  File "D:\blender\blender-2.80.0-git.85322737baec-windows64\2.80\scripts\modules\bpy\ops.py", line 200, in __call__
    ret = op_call(self.idname_py(), None, kw)
RuntimeError: Error: Traceback (most recent call last):
  File "D:\git_pipeline\avalon\development\git\pyblish-qml\pyblish_qml\host.py", line 659, in execute
    self._timer = wm.event_timer_add(0.01, context.window)
TypeError: WindowManager.event_timer_add(): required parameter "window" to be a keyword argument!

location: D:\blender\blender-2.80.0-git.85322737baec-windows64\2.80\scripts\modules\bpy\ops.py:200
```
I guess the new Blender beta version is not supported in which I previously had it working without this code in QML to "support Blender".

### Blender 2.79

Then I installed Blender 2.79 - no luck on "just seeing the menu", but then maybe I just don't know how to 'install pyblish' for Blender. ;) Anyway, I ran `pyblish_qml.show()` and there it did open and seemed to run fine (even though I had no plug-ins running anything in Blender only some default collectors:

afbeelding


---

This was being tested with the `find_pyqt5` reverted that triggered a `subprocess`, so this commit was reverted: https://github.com/pyblish/pyblish-qml/commit/31870fbd920927562d7459fe9f69f1cfe0b45bf9

Here's the environment I used to run blender for both versions:
```
set PYTHONPATH=;D:\git\pyblish-qml;D:\git\pyblish-base;%PYTHONPATH%
set PYBLISH_QML_PYTHON_EXECUTABLE=D:\bin\windows\python36\python.exe
blender
```

mottosso:
Ah, yes you are right; my memory failed me. When I wrote this, I also wrote an (in-house) integration to handle the menu item. Calling `show` is all QML should provide, and if it works fine without this commit then that's all fine. Not sure why I wrote that in the first place, but it was admittedly in the heat of production.

Happy with this, and thanks for confirming!

BigRoy:
Cool - do note that Blender 2.8 support for QML then still is pending, but that should be a separate issue. Thanks! :)

davidlatwe:
Thanks @BigRoy !
If everyone is happy, I'll bump the version and merge tomorrow :D

PR #332, opened 2019-05-31
Fix redundant newline from #329

davidlatwe:
### Problem
Still getting redundant newline after #329.

### Fix
It looks like previous fix missed out to catch *regular message*, and stripping them solved the issue.
So this PR seems to be the last puzzle of #329.

### DIFF
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -259,7 +259,9 @@ class Server(object):
                         last_msg_newline = True
                     else:
                         # This must be a regular message.
-                        sys.stdout.write(line)
+                        line = line.strip()
+                        if line:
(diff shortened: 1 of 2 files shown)


BigRoy:
Thanks, the empty lines getting printed seem to be gone with this PR. 👍

pyblish_qml/ipc/server.py
```diff
@@ -259,7 +259,9 @@ def _listen():
                         last_msg_newline = True
                     else:
                         # This must be a regular message.
-                        sys.stdout.write(line)
+                        line = line.strip()
+                        if line:
```
mottosso: That's clever. :)

davidlatwe:
About to merge this :)

PR #347, opened 2019-10-19
Fix#335

### LINKED ISSUE #335: Supplying targets to show() does not refresh when window is already open
### Issue

The command `pyblish_qml.show(targets=["render"])` will not update the targets of the running Pyblish QML session if it is already open. It does appear to refresh (it restarts the Collecting stage) however it makes no changes whatsoever to the targets as long as the interface is open.

*Workaround*: Close the Pyblish QML interface and then run it with the explicit targets set.

To reproduce:

1. Run `pyblish_qml.show(targets=["default"])`
2. Wait for Pyblish QML interface to start and initialize.
3. Run `pyblish_qml.show(targets=["render"])`
4. The interface appears to reinitialize, however it still used the `default` target.

---

##### Also note: `pyblish.api.register_target` does not propagate to an open Pyblish QML session.

The live QML session does *not* whatsoever update to any registered targets, so say:

```python
pyblish.api.register_target("render")
# Now refresh the QML ui, no change...
```

However, I think this might be a deliberate **design choice** so the explicitly defined target on launch does not require the targets to be globally registered. As such, you can target the specific Pyblish QML session without changing the environment which is useful.

davidlatwe:
This PR resolves #335, with some PEP8 changes, and fixed one minor bug.

### DIFF
--- a/pyblish_qml/__init__.py
+++ b/pyblish_qml/__init__.py
@@ -5,7 +5,12 @@ from .version import (
 )
 
 
-def show(parent=None, targets=None, modal=None, foster=None, auto_publish=False, auto_validate=False):
+def show(parent=None,
+         targets=None,
+         modal=None,
+         foster=None,
+         auto_publish=False,
+         auto_validate=False):
     from . import host
 
     if foster is not None:
--- a/pyblish_qml/app.py
+++ b/pyblish_qml/app.py
@@ -182,8 +182,10 @@ class Application(QtGui.QGuiApplication):
             util.timer_end("ready", "Awaited statemachine for %.2f ms")
 
         if client_settings:
-            self.controller.data['autoValidate'] = client_settings.get('autoValidate', False)
-            self.controller.data['autoPublish'] = client_settings.get('autoPublish', False)
+            auto_validate = client_settings.get('autoValidate', False)
+            auto_publish = client_settings.get('autoPublish', False)
+            self.controller.data['autoValidate'] = auto_validate
+            self.controller.data['autoPublish'] = auto_publish
 
         self.controller.show.emit()
 
--- a/pyblish_qml/app.py
+++ b/pyblish_qml/app.py
@@ -70,6 +70,8 @@ class Application(QtGui.QGuiApplication):
     published = QtCore.pyqtSignal()
     validated = QtCore.pyqtSignal()
 
+    targeted = QtCore.pyqtSignal(QtCore.QVariant)
+
     risen = QtCore.pyqtSignal()
     inFocused = QtCore.pyqtSignal()
     outFocused = QtCore.pyqtSignal()
@@ -104,6 +106,8 @@ class Application(QtGui.QGuiApplication):
         self.quitted.connect(self.quit)
         self.published.connect(self.publish)
         self.validated.connect(self.validate)
+
+        self.targeted.connect(self.target)
 
         self.risen.connect(self.rise)
         self.inFocused.connect(self.inFocus)
@@ -225,6 +229,9 @@ class Application(QtGui.QGuiApplication):
         """Fire up the validation sequance"""
         self.controller.validate()
 
+    def target(self, targets):
+        self.controller.targets = targets
+
     def listen(self):
         """Listen on incoming messages from host
 
@@ -250,6 +257,8 @@ class Application(QtGui.QGuiApplication):
                     "quit": "quitted",
                     "publish": "published",
                     "validate": "validated",
+
+                    "target": "targeted",
 
                     "rise": "risen",
                     "inFocus": "inFocused",
--- a/pyblish_qml/host.py
+++ b/pyblish_qml/host.py
@@ -110,6 +110,8 @@ def show(parent=None,
         proxy = ipc.server.Proxy(server)
 
         try:
+            # Update target
+            proxy.target(targets)
             proxy.show(show_settings)
             return server
 
--- a/pyblish_qml/host.py
+++ b/pyblish_qml/host.py
@@ -76,7 +76,11 @@ def uninstall():
     sys.stdout.write("Pyblish QML shutdown successful.\n")
 
 
-def show(parent=None, targets=[], modal=None, auto_publish=False, auto_validate=False):
+def show(parent=None,
+         targets=[],
+         modal=None,
+         auto_publish=False,
+         auto_validate=False):
     """Attempt to show GUI
 
     Requires install() to have been run first, and
@@ -163,7 +167,7 @@ def validate(proxy):
     proxy.validate()
 
 
-@proxy_call 
+@proxy_call
 def hide(proxy):
     proxy.hide()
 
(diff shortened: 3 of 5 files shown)


davidlatwe:
Was thinking about adding buttons for switching targets directly in QML, and here's what I have got.

qml_direction

### Directions

The newly added tab was called `directions`.
I named it `directions` was because it's *a list of `targets`*, so I thought maybe we could say a list of target is a direction (like a vector), and this tab is listing all directions for user to pick.

For listing multiple `directions`, you have to define them in `pyblish_qml.settings`, like:
```python
pyblish_qml.settings.Directions = {
    "Remote Publish": {
        "description": "Publish in remote machine",
        "targets": ["default", "deadline"],
    },
    "Local Publish": {
        "description": "Publish via local machine",
        "targets": ["default", "localhost"],
    },
}
```

QML will pick up and listing them in tab as buttons, and indicate current targets.

Note that this is backward compatible, so it's okay to leave `pyblish_qml.settings` untouched, if you only use one set of targets.

### Motivation

After #335 been fixed, you still need to press button *out side* of QML for changing targets, so I though maybe we could improve this a bit further by adding this feature.

Please let me know what you think, and I'll elaborate more later today. :)

mottosso:
Looks interesting. I'm a little unsure of how targets are being used, so I can't give a definite opinion on how to represent it visually. Overall, I would have thought "target" is the most important and first choice to make when publishing. Like when you go to McDonalds and make your selection in one of their order screens, you start by choosing "Eat in" or "Take away".

For targets, I would have thought that this choice should come even before a Pyblish GUI is shown, like as another option for showing the GUI.

- Publish Local
- Publish Remote

With that in mind, it seems a little backwards to have it nested three or more clicks into the GUI. But then again, maybe this is a decision you make late, or several times from within the same publishing session?

> I named it directions was because it's a list of targets, so I thought maybe we could say a list of target is a direction (like a vector), and this tab is listing all directions for user to pick.

Not too sure about this one, would rather not introduce more jargon; I can already hear the questions in email, chat and forums xD

How about making it a list if checkboxes instead? One where `default` is already checked? Alternatively, do we need multiple targets per target? I.e. can we have just `localhost` and `deadline` and implicitly include `default`?

davidlatwe:
> Overall, I would have thought "target" is the most important and first choice to make when publishing.
...
> For targets, I would have thought that this choice should come even before a Pyblish GUI is shown

Yes, that's definitely, so in production there's a few buttons (currently two of them here) for artist to pick.

> maybe this is a decision you make late, or several times from within the same publishing session?

Yes ! That's what it's for, but also an indicator for current targets.

> How about making it a list if checkboxes instead? One where default is already checked?

I have thought about that too. :P

But I think the best is to make user can only select one target from GUI, since there may have conflict between targets. Like `local` and `remote`, it's obviously redundant to enable both of them. And may a bit hard to implement the targets' relation logic onto GUI, like you can have A and B but you can not have B and C at the same time.

> Alternatively, do we need multiple targets per target?

I think we need, `target` for me, may refer to *the site which actually handling the final publish process* and *the destination of the publish result*. Like `dailies`, it could work with `local` or `remote`.

tokejepsen:
What happens with the environment/plugins when you choose a different target?
The collection would need to be refreshed when choosing a different target.

> The newly added tab was called directions.

I think the terminology will get confusing if you call the tab `directions` rather than `targets`.

> Yes ! That's what it's for, but also an indicator for current targets.

@davidlatwe do you have a use-case for using different targets within the same session?

davidlatwe:
> What happens with the environment/plugins when you choose a different target?

If you only selecting targets, nothing will happen, the targets only get sets to host when you press reset button on and only on that tab.

> I think the terminology will get confusing if you call the tab directions rather than targets.

Haha, second objection vote ! Then I'll change that in GUI, but I think I still need that name internally for naming variables, need a think. :P

> do you have a use-case for using different targets within the same session?

The example use case would be, like publishing cameras and renders from the same scene, the camera only need to be published from local machine, but the renders require to be sent to render farm. So there's two sets of targets need to switch. :relaxed:

tokejepsen:
> If you only selecting targets, nothing will happen, the targets only get sets to host when you press reset button on and only on that tab. And once you press the reset, yes, the collection would be refreshed as well. :)

I completely forgot how the targets worked there :)

Nice work!

BigRoy:
_On vacation till Wednesday so my reply might be brief._

Looks really nice!! Some artists on our end will love this.

Not sure about the `directions` name and having more terminology around, but allowing to switch within Pyblish I think some Artists on our end might love quite a bit. It's like exposed allowed combinations of `targets`. Like a targets group or targets combo. What @mottosso describes as "checkboxes" might be problematic on our end as some combinations of targets should not be used together, e.g. `remote` with `local`. So I can see how exposing allowed `targets` groups would be useful.

I wonder if we need that separate page or whether it could be part of something like a pop-up combobox styled like that (very nice and clear!) when maybe clicking on the targets or having a clear button on the main page that describes you can switch. In the meantime the standalone page is a very nice setup.

davidlatwe:
Thanks @BigRoy. :D

Sorry again for not posting that new GUI stuff in another issue. A bit excited at the time :/
I'll push the code in a new PR.

Back to the main issue on fixing #335, I think we could merge this ?

BigRoy:
Unable to test this myself. But just the fix itself for changing the targets in the main session would be great to see merged. Other than that, the GUI issue into its own PR would make things easier yes. Nice work on both though!

mottosso:
> Back to the main issue on fixing #335, I think we could merge this ?

Hold up a minute, at the moment, `targets` is being changed by calling on the `Application`, which modifies the contained `Controller`. That does work, but it looks like a really roundabout way of accomplishing it. Are there any alterantives to this? Can we reach the controller, or Pyblish running in the host, some other way?

davidlatwe:
> Are there any alterantives to this? Can we reach the controller, or Pyblish running in the host, some other way?

Hmmm, I don't think so. Since the main pyblish logic ([main iter](https://github.com/pyblish/pyblish-qml/blob/d5fbb73dc15228573e504857df4e84ec18904ff0/pyblish_qml/control.py#L320)) is running in `Controller`, and filtering plugins by targets happens [here](https://github.com/pyblish/pyblish-qml/blob/d5fbb73dc15228573e504857df4e84ec18904ff0/pyblish_qml/control.py#L848) (also in `Controller`).

So the host only responsible for discovering plugins and executing plugins which assigned by `Controller`, based on that, I think contact the `Application` is the only way for now. :thinking:

davidlatwe:
The `targets` changing workflow after this PR is as follow :

1. Firing new targets **while QML still open** by `pyblish_qml.show(targets=["new_target"])`
2. The host using `server.Proxy` to send new targets
3. `Application` receive new targets
4. `Application` put new targets to `Controller`
5. `Controller` reset

mottosso:
Ok, I think that makes sense. It runs via the same "channel" as showing/hiding the GUI, whilst still running. Was a little worried we introduced a new line of communication, when we clearly are already capable of communicating with plug-ins/instances. But this looks legit, thanks for confirming.

Happy with this. 👍

davidlatwe:
Thanks @mottosso :relaxed: 
Preparing merge now.

PR #349, opened 2019-10-21
Python37support

### LINKED ISSUE #333: Demo Crashes
Environment
OS : Windows 10
Python : 3.7 (64-bit)

I tried to run the pyblish-qml demo with "python.exe - m pyblish_qml --demo" .
The window pops up with a bunch of demo validation checks.
But when I click the "publish" button in the lower right corner (white triangle),
it crashes python.
Apparently a python object was expected to be a dictionary, but wasn't. 

Here is output info from the command line 

"C:\python37-64\python.exe -m pyblish_qml --demo"
Starting pyblish-qml server..
Using Python @ 'C:\python37-64\python.exe'
Using PyQt5 @ 'C:\python37-64\Lib\site-packages\pyqt5'
Targets: default
Starting pyblish-qml
Done, don't forget to call `show()`
Entering state: "hidden"
Entering state: "ready"
Entering state: "clean"
Entering state: "alive"
Settings:
WindowSize = [430, 600]
HiddenSections = ['Collect']
HeartbeatInterval = 60
WindowPosition = [100, 100]
ContextLabel = Context
WindowTitle = Pyblish
Entering state: "visible"
Entering state: "initialising"
Entering state: "collecting"
Spent 3850.61 ms resetting
Made 0 requests during publish.
Spent 4165.39 ms resetting
Made 0 requests during reset.
Entering state: "ready"
No local comment, reading from context..
Entering state: "validating"
Entering state: "extracting"

File "C:\python37-64\Lib\site-packages\pyblish_qml\control.py", line 1004, in on_next
self.data["models"]["item"].update_with_result(result)

File "C:\python37-64\Lib\site-packages\pyblish_qml\models.py", line 484, in update_with_result
assert isinstance(result, dict), "%s is not a dictionary" % result

AssertionError: generator raised StopIteration is not a dictionary

mottosso:
Also fixes #333, also see https://github.com/pyblish/pyblish-base/pull/355

I'm not using this in production at the moment, so keep an eye out for anything you find missing. I suspect the GUI might have trouble displayin the message previously passed on from StopIteration, such as when validation failed.

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -1,6 +1,7 @@
 
 import json
 import time
+import logging
 import collections
 
 # Dependencies
@@ -11,6 +12,7 @@ import pyblish.logic
 from . import util, models, version, settings
 
 qtproperty = util.pyqtConstantProperty
+log = logging.getLogger(__name__)
 
 
 class Controller(QtCore.QObject):
@@ -351,11 +353,11 @@ class Controller(QtCore.QObject):
                     signals.pop(order).emit()
 
             if not self.data["state"]["is_running"]:
-                raise StopIteration("Stopped")
+                return log.info("Stopped")
 
             if test(**state):
                 self.data["state"]["testPassed"] = False
-                raise StopIteration("Stopped due to %s" % test(**state))
+                return log.error("Stopped due to %s" % test(**state))
             self.data["state"]["testPassed"] = True
 
             try:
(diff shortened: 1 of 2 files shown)


pyblish_qml/control.py
```diff
@@ -1016,7 +1018,7 @@ def run(self, plugins, context, callback=None, callback_args=[]):
         # the GUI and commence next task.
         def on_next(result):
             if isinstance(result, StopIteration):
-                return on_finished(str(result))
+                return on_finished()
```
davidlatwe: I found that if we keep this line unchanged, and instead of changing `raise StopIteration` to `return log.error`, just `return StopIteration` will fix for Py37 and keep the message showing properly on GUI !

davidlatwe: Oops, just found out that this has been merged :P
I'll submit another PR then :)

mottosso: Great news 👍

PR #350, opened 2019-10-22
Fix GUI message prompting after #349

davidlatwe:
GUI message prompting like showing "validated", "failed validation", and "published with error" was broken after #349, this PR resolves that side-effect. :relaxed: 

Hooray on Python3.7 supported :tada:

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -353,11 +353,12 @@ class Controller(QtCore.QObject):
                     signals.pop(order).emit()
 
             if not self.data["state"]["is_running"]:
-                return log.info("Stopped")
+                return stopiter_by_demand("Stopped")
 
             if test(**state):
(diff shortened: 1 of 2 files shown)


pyblish_qml/control.py
```diff
@@ -1197,7 +1198,7 @@ def iterator(plugins, context):
 
         message = test(**state)
         if message:
-            return log.error("Stopped due to %s" % message)
+            return stopiter_by_error("Stopped due to %s" % message)
```
mottosso: Nah, let's just return StopIteration. Do we still need the log message if StopIteration works?

davidlatwe: Okay ! Remove the bandage !

mottosso:
Looking good, welcome to version up and release! Thanks @davidlatwe

PR #354, opened 2020-02-03
Opt-in PySide2 via Qt5.py

### LINKED ISSUE #301: Doesn't work with Nuke
As reported by Jakub over [in the chat](https://gitter.im/getavalon/Lobby?at=5bab3cec41177e0bc7aab6bf), there's something preventing Pyblish QML from working with Nuke.

### LINKED ISSUE #313: Pyblish QML failing to load from Maya after loading Houdin Engine plug-in
### Issue

Under **Windows 10** it seems that whenever the current Maya session has loaded the Houdini Engine plug-in that Pyblish QML fails to run afterwards. This is due to a DLL error on loading PyQt5. Even after unloading the Houdini Engine plug-in the problem remains for that Maya session, whenever Houdini has been active Pyblish QML will fail to run.

```python
# Running in Maya after Houdini Engine loaded...
import pyblish_qml
pyblish_qml.show()

# Output below:
Using Python @ 'D:/test/python36\python.exe'
Using PyQt5 @ 'None'
Targets: default
Success. QML server available as pyblish_qml.api.current_server()
Event filter has been installed.
  File "D:\test\python36\lib\runpy.py", line 193, in _run_module_as_main
ImportError: DLL load failed: The specified procedure could not be found.
    from PyQt5 import QtCore, QtGui, QtQuick, QtTest
  File "D:\git_pipeline\avalon\development\git\pyblish-qml\pyblish_qml\app.py", line 12, in <module>
    from . import app
  File "D:\git_pipeline\avalon\development\git\pyblish-qml\pyblish_qml\__main__.py", line 6, in <module>
    exec(code, run_globals)
  File "D:\test\python36\lib\runpy.py", line 85, in _run_code
    "__main__", mod_spec)
```

### To reproduce

Reproducible on: Windows 10 + Maya 2018.5 + Houdini 17.0.416

_We have been unable to reproduce the error on Windows 7 with the same Maya and Houdini - thus it seems to be related to something in Windows 10!_

The basis of the error is also reproducible by doing:

```python
import subprocess
subprocess.Popen("python")
```

In the opened python console (assuming that's also the Python interpreted used to run the QML in background) if you just do:

```python
from PyQt5 import QtWidgets
# DLL error
```

### What I've tried..

Trying to strip the environment of anything related to Houdini/Qt/Maya as I run the subprocess command that reproduces the DLL error. Didn't work for me.

Here's some example code I tried:
```python
import os
import subprocess
from subprocess import PIPE
import pprint

# Clean the environment for the subprocess to be safe... (doesn't help)
# Note that loading Houdini Engine does not change the Maya environment (before/after is exactly the same, perfect match)
env = os.environ.copy()
for key in env.copy():
    if any(key.lower().startswith(x) for x in ["qt", "maya", "houdini", "vray", "xgen", "redshift", "pyblish", "mtoa", "bifrost", "avalon", "atf", "arnold", "adsk", "colorbleed", "deadline", "substance", "pymel", "mash", "fbx", "ffmpeg", "com", "session"]):
        env.pop(key)
        continue
    
    value = env[key]
    for name in ["maya", "houdini", "python27", "mtoa"]:
        if name in value.lower():
            paths = value.split(";")
            paths = [p for p in paths if name not in p.lower()]
            value = ";".join(paths)
            env[key] = value
        
# Remove some random variables if they are there - just to be sure and keep as clean as possible env
env.pop("__APP__", None)
env.pop("__EXE__", None)
env.pop("__COMPAT_LAYER", None)
env.pop("PROMPT", None)
env.pop("ONEDRIVE", None)
env.pop("SCHEMA", None)

# Debug print the environment
pprint.pprint(env)

# Running it with this environment, doesn't work. 
subprocess.Popen("python", env=env)

# Or (other test) 
# Even if I overkill detach the process and run Python through shell - same issue.
# Using Process Explorer this way it's not a 'child' of the original Maya process
# so I thought maybe that would behave different, but didn't fix it either.
CREATE_NEW_PROCESS_GROUP = 0x00000200
DETACHED_PROCESS = 0x00000008
subprocess.Popen("python",
                 env=env, shell=True, stdin=PIPE, stdout=PIPE, stderr=PIPE,
                 creationflags=DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP)
```

Or instead of stripping from current environment, just take the SYSTEMROOT (so Python can actually run)

```python
import os
import subprocess
from subprocess import PIPE
import pprint

# Take only speciifc environment keys
env = {}
env["SYSTEMROOT"] = os.environ["SYSTEMROOT"]
env["PATH"] = ""  # need at least empty PATH in environment for PyQt5 to import (see __init__.py of PyQt5)
    
# Debug print the environment
pprint.pprint(env)

# Running it with this environment, doesn't work. 
subprocess.Popen("python", env=env)

# Or (other test) 
# Even if I overkill detach the process and run Python through shell - same issue.
# Using Process Explorer this way it's not a 'child' of the original Maya process
# so I thought maybe that would behave different, but didn't fix it either.
CREATE_NEW_PROCESS_GROUP = 0x00000200
DETACHED_PROCESS = 0x00000008
subprocess.Popen("python",
                 env=env, shell=True, stdin=PIPE, stdout=PIPE, stderr=PIPE,
                 creationflags=DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP)

# Or using close_fds = True
CREATE_NEW_PROCESS_GROUP = 0x00000200
DETACHED_PROCESS = 0x00000008
subprocess.Popen("python",
                 env=env,
                 close_fds=True,
                 creationflags=CREATE_NEW_PROCESS_GROUP)
```

Again, no luck. 

#### Whatever does fix it (manually)

**However!** when I open [Process Explorer](https://filehippo.com/download_process_explorer/), right click the `python.exe` process and press "Restart" it reopens the Python with the exact same environment... and voila, PyQt5 can import correctly.

It seems as if somehow Windows 10 is trying to use the DLL from Memory that was loaded in connection with the Houdini (in Maya) process making it fail to load the actual correct one for the Python process.

Any ideas on how this could be resolved?

---

Reference discussion on Pyblish Gitter:
- https://gitter.im/pyblish/pyblish?at=5c3382c9aac7082e6feb1117
- https://gitter.im/pyblish/pyblish?at=5c2c88cadb5b5c68834396ef
- https://gitter.im/pyblish/pyblish?at=5c34adf7357dd7655d377726

Thanks to @davidlatwe for also confirming the issue on Windows 10 and being unable to reproduce it on Windows 7. Appreciate the time you took to investigate, cheers!

davidlatwe:
This will fix #313 and fix #301.

Both #313 and #301 were issues that having `PyQt5` *DLL load failed* error on App startup, and through the tests what @BigRoy has done in #313, looks like it's a problem which possibly came from DCC host about how they load those Qt DLLs, and is hard to solve directly from our end.

So since those DCC were holding DLL which was from `PySide2`, making `PySide2` also usable for `pyblish_qml` could be a good workaround.

### Changes
* Make QML App able to use `PySide2` as Qt binding with the help from `Qt5.py`.
* Vendorized `Qt5.py`.

### Caveats (Resolved, see comment below)
May lose font when using `PySide2`.

There's a strange problem after using `PySide2`, is the QML's comment box may display weird characters on first run.

fontfamily

But if we don't assign `font.family` in QML [here](https://github.com/pyblish/pyblish-qml/blob/c8fea9c39d0b47b224fc0e522e1b01fdaeafcb7f/pyblish_qml/qml/CommentBox.qml#L66), the text will display correctly (with out custom font of course).

### DIFF
--- a/pyblish_qml/models.py
+++ b/pyblish_qml/models.py
@@ -128,7 +128,7 @@ class PropertyType(type(QtCore.QObject)):
 
             attrs[key + "Changed"] = notify
             attrs[key] = QtCore.Property(
-                type(value) if value is not None else QtCore.QVariant,
+                type(value) if value is not None else "QVariant",
                 fget=lambda self, k=key: getattr(self, cls.prefix + k, None),
                 fset=set_data(key, value),
                 notify=notify)
@@ -247,7 +247,7 @@ class AbstractModel(QtCore.QAbstractListModel):
             except Exception:
                 pass
 
-        return QtCore.QVariant()
+        return Item()
 
     def roleNames(self):
         return {
@@ -309,7 +309,7 @@ class ItemModel(AbstractModel):
 
         self.endResetModel()
 
-    @QtCore.Slot(QtCore.QVariant)
+    @QtCore.Slot("QVariant")
     def add_plugin(self, plugin):
         """Append `plugin` to model
 
@@ -381,7 +381,7 @@ class ItemModel(AbstractModel):
         item = self.add_item(item)
         self.plugins.append(item)
 
-    @QtCore.Slot(QtCore.QVariant)
+    @QtCore.Slot("QVariant")
     def add_instance(self, instance):
         """Append `instance` to model
 
@@ -445,7 +445,7 @@ class ItemModel(AbstractModel):
 
         return item
 
-    @QtCore.Slot(QtCore.QVariant)
+    @QtCore.Slot("QVariant")
     def add_context(self, context, label=None):
         """Append `context` to model
 
--- a/pyblish_qml/models.py
+++ b/pyblish_qml/models.py
@@ -246,7 +246,7 @@ class AbstractModel(QtCore.QAbstractListModel):
             except Exception:
                 pass
 
-        return Item()
+        return "QVariant"
 
     def roleNames(self):
         return {
--- a/pyblish_qml/models.py
+++ b/pyblish_qml/models.py
@@ -2,7 +2,7 @@ import re
 import time
 import logging
 
-from PyQt5 import QtCore
+from Qt5 import QtCore
 
 from . import util, settings
 from .vendor import six
@@ -117,7 +117,7 @@ class PropertyType(type(QtCore.QObject)):
             if key.startswith("__"):
                 continue
 
-            notify = QtCore.pyqtSignal()
+            notify = QtCore.Signal()
 
             def set_data(key, value):
                 def set_data(self, value):
@@ -127,7 +127,7 @@ class PropertyType(type(QtCore.QObject)):
                 return set_data
 
             attrs[key + "Changed"] = notify
-            attrs[key] = QtCore.pyqtProperty(
+            attrs[key] = QtCore.Property(
                 type(value) if value is not None else QtCore.QVariant,
                 fget=lambda self, k=key: getattr(self, cls.prefix + k, None),
                 fset=set_data(key, value),
@@ -145,7 +145,7 @@ class AbstractItem(QtCore.QObject):
 
     """
 
-    __datachanged__ = QtCore.pyqtSignal(QtCore.QObject)
+    __datachanged__ = QtCore.Signal(QtCore.QObject)
 
     def __str__(self):
         return self.name
@@ -199,7 +199,7 @@ class AbstractModel(QtCore.QAbstractListModel):
         super(AbstractModel, self).__init__(parent)
         self.items = util.ItemList(key="id")
 
-    @QtCore.pyqtSlot(int, result=QtCore.QObject)
+    @QtCore.Slot(int, result=QtCore.QObject)
     def item(self, index):
         return self.items[index]
 
@@ -309,7 +309,7 @@ class ItemModel(AbstractModel):
 
         self.endResetModel()
 
-    @QtCore.pyqtSlot(QtCore.QVariant)
+    @QtCore.Slot(QtCore.QVariant)
     def add_plugin(self, plugin):
         """Append `plugin` to model
 
@@ -381,7 +381,7 @@ class ItemModel(AbstractModel):
         item = self.add_item(item)
         self.plugins.append(item)
 
-    @QtCore.pyqtSlot(QtCore.QVariant)
+    @QtCore.Slot(QtCore.QVariant)
     def add_instance(self, instance):
         """Append `instance` to model
 
@@ -445,7 +445,7 @@ class ItemModel(AbstractModel):
 
         return item
 
-    @QtCore.pyqtSlot(QtCore.QVariant)
+    @QtCore.Slot(QtCore.QVariant)
     def add_context(self, context, label=None):
         """Append `context` to model
 
@@ -604,7 +604,7 @@ class ItemModel(AbstractModel):
 
 class ResultModel(AbstractModel):
 
-    added = QtCore.pyqtSignal()
+    added = QtCore.Signal()
 
     def add_item(self, item):
         item_ = defaults["result"].copy()
@@ -744,21 +744,21 @@ class ProxyModel(QtCore.QSortFilterProxyModel):
 
         self.setFilterCaseSensitivity(QtCore.Qt.CaseInsensitive)
 
-    @QtCore.pyqtSlot(int, result=QtCore.QObject)
+    @QtCore.Slot(int, result=QtCore.QObject)
     def item(self, index):
         index = self.index(index, 0, QtCore.QModelIndex())
         index = self.mapToSource(index)
         model = self.sourceModel()
         return model.items[index.row()]
 
-    @QtCore.pyqtSlot(str, result=QtCore.QObject)
+    @QtCore.Slot(str, result=QtCore.QObject)
     def itemByName(self, name):
         model = self.sourceModel()
         for item in model.items:
             if name == item.name:
                 return item
 
-    @QtCore.pyqtSlot(str, str)
+    @QtCore.Slot(str, str)
     def add_exclusion(self, role, value):
         """Exclude item if `role` equals `value`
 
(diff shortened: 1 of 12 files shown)


pyblish_qml/models.py
```diff
@@ -247,7 +246,7 @@ def data(self, index, role=QtCore.Qt.DisplayRole):
             except Exception:
                 pass
 
-        return QtCore.QVariant()
+        return Item()
```
davidlatwe: A little unsure about this one, since `PySide2` does not have `QVariant`, I replace it with `Item` because the `self.items` was containing objects generated from `Item()`.

But since it did not cause any error when I deliberately raising Exception in that try-except scope for testing, so I think it's legit.

mottosso: Hm, did you try returning `None` instead? Or even `"QVariant"`? QML would interpret QVariant as an invalid item, but may interpret `Item()` as a valid one, when it isn't?

mottosso:
Stellar work David. Wonder why the tests are failing on import? :O They throw a module not found error, which is a subclass of import error, which is being handled.

For the font, that's an odd one. Maybe best not to assign a font to that one textfield for the time being. Odds are there's an issue loading the font from disk. Could be path related, or even OpenGL related.

We could possibly work around it by packing the font into a Qt Resource instead, but it can wait.

pyblish_qml/models.py
davidlatwe: > did you try returning None instead? Or even "QVariant"? 

I haven't ! And I just did, looks like they all works without error, just displaying stuff incorrectly.

image

I'll change it to `"QVariant"` for consistency.

davidlatwe:
> Wonder why the tests are failing on import? :O They throw a module not found error, which is a subclass of import error, which is being handled.

Hmmm, it looks like it will raise eventually, [here](https://github.com/mottosso/Qt5.py/blob/f130087774be942ce0451501762277b69f3dedda/Qt5.py#L72).
Looks like we need to add `python3-pyqt5.qtmultimedia` this package into `Dockerfile` for test ?

pyblish_qml/models.py
davidlatwe: Changed in commit 14b58f9.

davidlatwe:
I have found the soultion for weird font appearance !

So it looks like PySide's font loading is a bit different from PyQt, using `FontLoader` to load font explicitly and reference it on demand solved the issue !

Also for the visual consistency, I set the `textRenderType` to `NativeTextRendering` because it's what I got from PyQt at my end before. Let me know if this doesn't work for you.

davidlatwe:
Also, the CI test was fixed in commit 9205fdd, and sorry for the force push, was trying to keep this kind of commit minimum.

davidlatwe:
> Also for the visual consistency, I set the textRenderType to NativeTextRendering because it's what I got from PyQt at my end before. Let me know if this doesn't work for you.

In fact, PyQt was rendering text in native way only because those texts (local chars) was not found in `"Open Sans"`. I should revert that change.

Should make font family be customizable, but yeah, it could wait till next PR.

mottosso:
Once you're happy with this @davidlatwe you're welcome to merge and release. Great work. 👍

davidlatwe:
So I have found and solved one last bug after this PR, it's a random error that raised from QML on window close. This seems only happens when using `PySide2` as Qt binding.

Not every time, but often, got these errors when I close the window:
```
file:///.../pyblish_qml/qml/Footer.qml:50: TypeError: Cannot read property 'hasComment' of null
file:///.../pyblish_qml/qml/Terminal.qml:144: TypeError: Cannot read property 'resultModel' of null
file:///.../pyblish_qml/qml/Terminal.qml:39: TypeError: Cannot read property 'resultProxy' of null
file:///.../pyblish_qml/qml/Overview.qml:117: TypeError: Cannot read property 'pluginProxy' of null
file:///.../pyblish_qml/qml/Overview.qml:84: TypeError: Cannot read property 'instanceProxy' of null
file:///.../pyblish_qml/qml/Overview.qml:125: TypeError: Cannot read property 'pluginProxy' of null
file:///.../pyblish_qml/qml/Overview.qml:125: TypeError: Cannot read property 'pluginProxy' of null
file:///.../pyblish_qml/qml/Overview.qml:92: TypeError: Cannot read property 'instanceProxy' of null
file:///.../pyblish_qml/qml/Overview.qml:125: TypeError: Cannot read property 'pluginProxy' of null
file:///.../pyblish_qml/qml/Overview.qml:125: TypeError: Cannot read property 'pluginProxy' of null
file:///.../pyblish_qml/qml/Perspective/Viewport.qml:31: TypeError: Cannot read property 'errorProxy' of null
```

QML was complaining about reading properties from a null object, and that object was a root context property called `app` which exposed from Python, an instance of `pyblish_qml.control.Controller`.

And it became null in QML was because it's garbage collected before QML component gets destroyed on window close, and that's why the error happens randomly.

The solution was simple, just assign a parent to the `pyblish_qml.control.Controller`, see commit 900d739.

Since the error happens randomly in our case, here's a [small example](https://gist.github.com/davidlatwe/44d6236570a11ebb869682a8d0b4f661) that proves the point.

Finally, will merge this tomorrow if no other objections. ☺️

mottosso:
> The solution was simple, just assign a parent to the pyblish_qml.control.Controller

Makes sense. :) Good find. Happy to merge when you're happy.

BigRoy:
@davidlatwe just wondering. Do I need to do anything to make this work inside Houdini, like an environment variable? Or would this magically work already?

I'm expecting I need to change something but I'm not sure what.

davidlatwe:
> Do I need to do anything to make this work inside Houdini

If you mean in Maya with Houdini Engine loaded, like the case in #313, then all you need is installing `PySide2` in the place that Python could find.

If not, Houdini should works just fine without doing anything else. :)

PR #355, opened 2020-02-03
Fix registered targets lost when re-launching on existed server

davidlatwe:
This was reported in Gitter 👉 [February 3, 2020 4:54 PM](https://gitter.im/pyblish/pyblish?at=5e37dfb6fe0e6f74e9fbd502) from @BigRoy .

It was a bug after #347, it was trying to solve #335 which was assigning and changing targets via `pyblish_qml.api.show`. But the solution did not pay attention to the case that is using targets from `pyblish.api.registered_targets`, hence we got the bug.

### Reproducible
```python
import pyblish.api
import pyblish_qml.api

# Register some random target.
pyblish.api.register_target("custom")
# Launch !
pyblish_qml.api.show()
# Now navigate to the `Terminal` page on top of GUI, should see our target "custom" in view.
# While window still exists, launch again.
pyblish_qml.api.show()
# Targets has been wiped..

```

### What's changed
* Only updating targets to current server if `targets` is provided
* Default value of `targets` changes to `None` instead of mutable `list`

### DIFF
--- a/pyblish_qml/app.py
+++ b/pyblish_qml/app.py
@@ -76,7 +76,7 @@ class Application(QtGui.QGuiApplication):
     inFocused = QtCore.pyqtSignal()
     outFocused = QtCore.pyqtSignal()
 
-    def __init__(self, source, targets=[]):
+    def __init__(self, source, targets=None):
         super(Application, self).__init__(sys.argv)
 
         self.setWindowIcon(QtGui.QIcon(ICON_PATH))
@@ -281,7 +281,7 @@ class Application(QtGui.QGuiApplication):
         thread.start()
 
 
-def main(demo=False, aschild=False, targets=[]):
+def main(demo=False, aschild=False, targets=None):
     """Start the Qt-runtime and show the window
 
     Arguments:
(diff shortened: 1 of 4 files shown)


BigRoy:
Nice - thanks for looking into it so quickly.

This could be a tricky one though. What happens on:

```python
pyblish_qml.show(targets=["custom"]
# ... keep window open
pyblish_qml.show(targets=None)
```

Would we want the latter to resort to the default of using `[default] + api.registered_targets()` like on the initial run or to have it persist whatever it had. 

It seems most logical if it behaved the same to a complete refresh (close and run) with the same arguments. As such, should it on `targets=None` be `targets = [default] + api.registered_targets()` [as is done on `__init__`](https://github.com/pyblish/pyblish-qml/blob/71df17d701efeb74f128fc320350172ab4c1b20a/pyblish_qml/ipc/server.py#L204-L209). I'd expect that to be the most logical behavior. Does it currently behave like that?

davidlatwe:
> Does it currently behave like that?

No ! And I think you are right, that would be what I'll expected, too.

davidlatwe:
>Does it currently behave like that?

And now it does.

Previously, `targets` provided from `pyblish_qml.api.show` will remain the same if next `show` call has no `targets` given. After 3fdccb5, it will fallback to `[default] + pyblish.api.registered_targets()`.

davidlatwe:
Thanks ! Will merge this later today :D

davidlatwe:
Arrr sorry, forgot to bump the version, will push to master. 😭

PR #362, opened 2020-03-08
Fix error handling within iterator

### LINKED ISSUE #353: 'return' with argument inside generator
Whenever I run ```pyblish_qml.show()``` either from maya or ```python2 -m pyblish_qml --demo``` I successfully load the QML server, then am met with 
```
File "P:\util\maya\Maya2020\experimental\lib\Python27\lib\runpy.py", line 162, in _run_module_as_main
SyntaxError: 'return' with argument inside generator
yield result
File "P:\util\maya\Maya2020\experimental\pyblish_qml\control.py", line 378
from . import util, compat, control, settings, ipc
File "P:\util\maya\Maya2020\experimental\pyblish_qml\app.py", line 15, in <module>
from . import app
File "P:\util\maya\Maya2020\experimental\pyblish_qml\__main__.py", line 6, in <module>
exec code in run_globals
File "P:\util\maya\Maya2020\experimental\lib\Python27\lib\runpy.py", line 72, in _run_code
"__main__", fname, loader, pkg_name)
```

It is my understanding that this is behavior that works in python3 but not 2.7, which is a necessity for Maya. Do I need to check out a different branch for 2.7 support ?

buddly27:
This resolves #353

Tested with PySide2 and Python 2.7, Python 3.7 and Python 3.8 with the demo window:

```
python -m pyblish_qml --demo --targets default
```
(took me a while to figure out that I had to pass the 'default' target to see something!)

I didn't deal with unit tests as I am unfamiliar with nose, please let me know if you want me to handle it.

### DIFF
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -353,29 +353,24 @@ class Controller(QtCore.QObject):
                     signals.pop(order).emit()
 
             if not self.data["state"]["is_running"]:
-                return StopIteration("Stopped")
+                return
 
             if test(**state):
                 self.data["state"]["testPassed"] = False
-                return StopIteration("Stopped due to %s" % test(**state))
+                raise RuntimeError("Stopped due to %s" % test(**state))
 
             self.data["state"]["testPassed"] = True
 
-            try:
-                # Notify GUI before commencing remote processing
-                self.about_to_process.emit(plug, instance)
+            # Notify GUI before commencing remote processing
+            self.about_to_process.emit(plug, instance)
 
-                result = self.host.process(plug, context, instance)
+            result = self.host.process(plug, context, instance)
 
-            except Exception as e:
-                return StopIteration("Unknown error: %s" % e)
-
-            else:
-                # Make note of the order at which the
-                # potential error error occured.
-                has_error = result["error"] is not None
-                if has_error:
-                    state["ordersWithError"].add(plug.order)
+            # Make note of the order at which the
+            # potential error error occured.
+            has_error = result["error"] is not None
+            if has_error:
+                state["ordersWithError"].add(plug.order)
 
             yield result
 
@@ -1207,7 +1202,7 @@ def iterator(plugins, context):
 
         message = test(**state)
         if message:
-            return StopIteration("Stopped due to %s" % message)
+            raise RuntimeError("Stopped due to %s" % message)
 
         instances = pyblish.api.instances_by_plugin(context, plugin)
         if plugin.__instanceEnabled__:
--- a/pyblish_qml/control.py
+++ b/pyblish_qml/control.py
@@ -353,24 +353,32 @@ class Controller(QtCore.QObject):
                     signals.pop(order).emit()
 
             if not self.data["state"]["is_running"]:
+                yield StopIteration("Stopped")
                 return
 
             if test(**state):
                 self.data["state"]["testPassed"] = False
-                raise RuntimeError("Stopped due to %s" % test(**state))
+                yield StopIteration("Stopped due to %s" % test(**state))
+                return
 
             self.data["state"]["testPassed"] = True
 
-            # Notify GUI before commencing remote processing
-            self.about_to_process.emit(plug, instance)
+            try:
+                # Notify GUI before commencing remote processing
+                self.about_to_process.emit(plug, instance)
 
-            result = self.host.process(plug, context, instance)
+                result = self.host.process(plug, context, instance)
 
-            # Make note of the order at which the
-            # potential error error occured.
-            has_error = result["error"] is not None
-            if has_error:
-                state["ordersWithError"].add(plug.order)
+            except Exception as e:
+                yield StopIteration("Unknown error: %s" % e)
+                return
+
+            else:
+                # Make note of the order at which the
+                # potential error error occured.
+                has_error = result["error"] is not None
+                if has_error:
+                    state["ordersWithError"].add(plug.order)
 
             yield result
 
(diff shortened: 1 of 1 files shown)


pyblish_qml/control.py
```diff
@@ -353,29 +353,24 @@ def iterator(self, plugins, context):
                     signals.pop(order).emit()
 
             if not self.data["state"]["is_running"]:
-                return StopIteration("Stopped")
+                return
```
mottosso: I think we're missing a yield here, to preserve the message "Stopped".

```py
yield "Stopped"
return
```

pyblish_qml/control.py
```diff
@@ -353,29 +353,24 @@ def iterator(self, plugins, context):
...
-                # Notify GUI before commencing remote processing
-                self.about_to_process.emit(plug, instance)
-
-                result = self.host.process(plug, context, instance)
+            # Notify GUI before commencing remote processing
+            self.about_to_process.emit(plug, instance)
 
-            except Exception as e:
-                return StopIteration("Unknown error: %s" % e)
+            result = self.host.process(plug, context, instance)
```
mottosso: Is it safe to remove this try/except block? self.host.process can throw an error, in which case we aren't handling it properly. That's what the try/except was doing. Also, why are we removing it? Better do another `yield` and `return` here.

pyblish_qml/control.py
```diff
@@ -1162,7 +1157,7 @@ def on_next(result):
             if not self.data["state"]["is_running"]:
                 return on_finished()
 
-            if isinstance(result, StopIteration):
+            if isinstance(result, Exception):
```
mottosso: This probably won't work. Now, whenever there is *any* exception, including actual faults or bugs, it will think the iterator is finished and exit successfully.

StopIteration was only ever provided if the iterator finished successfully. If we can't return a StopIteration, we need to pass something else, to differentiate an error from a success.

pyblish_qml/control.py
```diff
@@ -1207,7 +1202,7 @@ def iterator(plugins, context):
 
         message = test(**state)
         if message:
-            return StopIteration("Stopped due to %s" % message)
+            raise RuntimeError("Stopped due to %s" % message)
```
mottosso: We probably can't raise here. The iterator used above expects to never fail and always yield a value. Now that it throws, we need a try/except to handle the case where it doesn't yield and instead throws an exception.

I think we should be able to instead..

```py
yield StopIteration("...")
return
```

That should replicate what it was already doing before this yielding thing became an issue.

mottosso (commented):
Thanks for this, there are a few things I suspect might cause issues, have a look at my comments.

pyblish_qml/control.py
buddly27: By removing the StopIteration exception, I considered that all messages that needed to be transferred were error messages which could be embedded into other type of Exception (e.g. RuntimeError), but this would indeed not work in this case as we wouldn't want to raise a RuntimeError just to get the "Stopped" message.

Yielding "Stopping" as you suggest would imply a check on the "on_next" callback to verify if "result" is a string, which is what I wanted to prevent.

So I think I'm gonna revert to simply apply the logic recommended in this comment instead:
https://github.com/pyblish/pyblish-qml/issues/353#issuecomment-584159312

pyblish_qml/control.py
buddly27: It is safe as all exception would be caught in the `util.defer` logic and transmitted to the `on_next` callback as an Exception object. However as you mentioned in your other [comment](https://github.com/pyblish/pyblish-qml/pull/362/files#r389347317), it means that all exceptions would be routed in the `on_finished` callback which is not great.

pyblish_qml/control.py
buddly27: Done

pyblish_qml/control.py
buddly27: Reverted

pyblish_qml/control.py
buddly27: Reverted

pyblish_qml/control.py
buddly27: Updated to use suggested syntax.

buddly27:
I reverted to using the logic suggested in https://github.com/pyblish/pyblish-qml/issues/353#issuecomment-584159312.

Removing usage of `StopIteration` would require more thoughts to prevent the issues you highlighted.

mottosso:
> (took me a while to figure out that I had to pass the 'default' target to see something!)

Do you mean that just calling `python -m pyblish_qml --demo` isn't enough to see anything, that you also need `--targets default`? :O That shouldn't be the case, might be a bug.

buddly27:
> Do you mean that just calling python -m pyblish_qml --demo isn't enough to see anything, that you also need --targets default? :O That shouldn't be the case, might be a bug.)

Yes that's what I mean. I guess it could be because targets default to `["default"]` in `pyblish_qml.host.show`, but not in `pyblish_qml.ipc.server.pyblish_qml`. Though I feel default targets should probably be handled by [pyblish-base](https://github.com/pyblish/pyblish-base) instead for a better separation of concern between UI and logic, don't you think?

https://github.com/pyblish/pyblish-qml/blob/18826c8cad919b61e2c5332055242f85db814fd6/pyblish_qml/host.py#L100

https://github.com/pyblish/pyblish-qml/blob/18826c8cad919b61e2c5332055242f85db814fd6/pyblish_qml/ipc/server.py#L206

mottosso:
> Yes that's what I mean.

Can confirm this is happening, it happened in the latest release 1.11.1 and works as intended in 1.11.0. I've pushed a fix for it now.

https://github.com/pyblish/pyblish-qml/commit/835bf3bcc210fb181392502159438237597698ec

That's beside the point in this PR however. Let me know when you're ready for another review.

buddly27:
> Can confirm this is happening, it happened in the latest release 1.11.1 and works as intended in 1.11.0. I've pushed a fix for it now.

Thanks!

> Let me know when you're ready for another review.

Ready now :)

buddly27:
Just merged back latest change to this branch, let me know if there are any issues still!

PR #366, opened 2020-06-02
Relax pyblish-base version dependency.

### LINKED ISSUE #337: Inconsistent pyblish-base requirements
The required version of `pyblish-base` is inconsistently specified both within this package and in the `pyblish` meta package:

Within `pyblish-qml` requirements.txt says:
`pyblish-base>=1.4`

while setup.py says:
`pyblish-base==1.5.3`

In addition, `pyblish` 1.6.0 requires:
`pyblish-base==1.7.2`
`pyblish-qml==1.8.5`

which are mutually incompatible because of the dependency on 1.5.3 in `pyblish-qml`.

jedfrechette:
Closes #337 

I'm not sure this actually sets the version dependency appropriately, but pyblish-qml almost certainly shouldn't be explicitly depending on such an old version of pyblish-base as it is now.

### DIFF
--- a/pyblish_qml/version.py
+++ b/pyblish_qml/version.py
@@ -1,7 +1,7 @@
 
 VERSION_MAJOR = 1
 VERSION_MINOR = 11
-VERSION_PATCH = 3
+VERSION_PATCH = 4
 
 version_info = (VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH)
 version = '%i.%i.%i' % version_info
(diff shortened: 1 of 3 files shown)


mottosso:
Mm, yes that's true. This will at least not fail or require you to uninstall a newever version to install pyblish-qml. This looks good to me. Could you version up as well, in `version.py`? That way I can make a new release with this change.

mottosso:
Released, thanks @jedfrechette!

PR #373, opened 2022-06-27
add sys.path to python path

hannesdelbeke:
this PR adds support to the QML server working in certain isolated environments
without having to do any additional custom env management

follow up from the gitter discussion:

Hannes @hannesdelbeke
> in qml, (server.py) we pass the absolute minimum of environment variables to the server, to avoid issues on invalid types
> 
> we do pass PYTHONPATH.,
> 
> ```python
> # Append PyQt5 to existing PYTHONPATH, if available
>         environ["PYTHONPATH"] = os.pathsep.join(
>             path for path in [os.getenv("PYTHONPATH"), pyqt5]
>             if path is not None
>         )
> ```
> but we don't pass sys.path.
> 
> so python modules accessible to our software (ex. blender), but which are not in the PYTHONPATH, are thus not accessible to our server.
> 
> which means QML cant run if it's installed in your user script path.
> 
> passing sys.path fixes the issue. is this too specific to create a PR for?
> ```python
> # Append sys.path to PYTHONPATH
>         if environ["PYTHONPATH"] is not None:
>             environ["PYTHONPATH"] += os.pathsep + os.pathsep.join(sys.path)
>         else:
>             environ["PYTHONPATH"] = os.pathsep.join(sys.path)
> ```

Marcus Ottosson @mottosso May 20 06:44
> Heya, before you make a PR, just to clarify a common misconception about QML; the PYTHONPATH used by QML are unrelated to your collector/validator plug-ins. They only relate to finding the PyQt5/PySide2 to draw the UI. Plug-ins are still run in your local environment, with your local sys.path

Hannes @hannesdelbeke May 20 10:24

> yes the issue is not with the plugins. those all work fine.
> 
> the server needs to import the QML module.
> if pyblish_qml is not in PYTHONPATH, but only in your sys.path, it will fail to launch qml.

### DIFF
--- a/pyblish_qml/ipc/server.py
+++ b/pyblish_qml/ipc/server.py
@@ -157,6 +157,12 @@ class Server(object):
             path for path in [os.getenv("PYTHONPATH"), pyqt5]
             if path is not None
         )
+        
+        # Append sys.path to PYTHONPATH
+        if environ["PYTHONPATH"] is not None:
+            environ["PYTHONPATH"] += os.pathsep + os.pathsep.join(sys.path)
+        else:
+            environ["PYTHONPATH"] = os.pathsep.join(sys.path)
 
         # Protect against an erroneous parent environment
         # The environment passed to subprocess is inherited from


mottosso:
The design intent is for QML to have its own dedicated Python, where QML is already installed into its own site-packages. A separate, completely independent Python typically located on the network which is only ever capable and meant to be run with Pyblish QML. In that case, you don't need any variables, just set the path to that executable and it'll know where to find QML.

Anything else is an optimisation for 50 mb of disk space and increases the complexity of your setup.

PR #380, opened 2024-06-08
Improve compatibility with Python 3.10 and newer

jedfrechette:


mottosso:
Brilliant, thanks for the contribution @jedfrechette

PR #381, opened 2024-08-01
Fix Python 3 getfullargspec attribute name compatibility

jedfrechette:
Fix another issue due to the removal of getargspec. Although returning the same info, some attribute names have changed with the move to getfullargspec, I've updated the active code where this was an issue and removed the vendorized mock.py, which will have similar issues but is not used anywhere and is badly outdated.


mottosso:
Thanks for this, sorry for the delay. I struggle to review Pyblish-related code as I'm not using it myself. Would you like to handle merge requests and releases moving forward?

jedfrechette:
> Would you like to handle merge requests and releases moving forward?
I can try to help, but a lot of pyblish-qml's internals look a bit like magic to me so I would have trouble evaluating any complex changes.

mottosso:
Cool, I'll merge this now, keep an eye out for any issues.

PR #383, opened 2024-08-05
add a new host for UnrealEngine

SKY123C:
i has create a pr for pyblish-unrea, .i creates an executeInMainThreadWithResult func for unreal module in this plugin


mottosso:
Very cool, double-merged!

